PR #2, opened 2019-12-16
More style changes

gebner:
* Make the instances list collapsible.
 * Show constructors and fields in Lean syntax.

more_style_changes


robertylewis:
Merged!

PR #3, opened 2019-12-17
Link notations, projections, fields in structure instances

gebner:
This requires the companion PR leanprover-community/lean#89.

doc_links

(The notation for `∩` is not linked since `has_inter.inter` is not supported yet by the linkification function.)

I think we should probably disable the underlines from the links though.


gebner:
With Rob's latest changes, even the field notation is linked!

PR #5, opened 2019-12-17
declaration names link to anchor; github links at bottom

robertylewis:
The declaration names now link to their own position on the page.

Maybe the spacing needs to be adjusted a little, I'm not sure. (Or we could use a normal bullet point.)

gh_link

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -214,7 +214,7 @@ search_snippet = """
 def write_internal_nav(objs, filename, out):
   out.write('<h1>Lean <a href="https://leanprover-community.github.io">mathlib</a> docs</h1>')
   out.write('<h2><a href="#top">{0}</a></h2>'.format(filename_import(filename)))
-  out.write('<div class="gh_link"><a href="{}">(view source on GitHub)</a></div>'.format(library_link(filename)))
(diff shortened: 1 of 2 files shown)


gebner:
I also found it unintuitive that the declaration names linked to the source, unlike all other names.  The explicit `View declaration on GitHub` link on every declaration is IMHO quite distracting.  Every theorem now has a second line that conveys no information.

My first reaction to reduce the visual clutter would be to move the github link to the right:
doc_view_source

https://github.com/gebner/doc-gen/commit/6c50c3b77ae107fe5e15ddecd52fa4c1ab4ddbbf

robertylewis:
I agree putting it at the bottom isn't ideal. On a wide screen, floating it to the right kind of disassociates it from the declaration, especially for declarations that don't have many arguments.
Screenshot from 2019-12-18 16-08-09
Maybe repeating the colored line on the right of the declaration div would improve it? I'm not sure -- I'll experiment a bit.

robertylewis:
How about this?
Screenshot from 2019-12-18 16-33-42

gebner:
I think the horizontal line looks good.  However putting borders on all sides looks a bit too heavy to me.  How about just putting borders on the left & top, like this?

horizontal_line_proposal1

robertylewis:
I like it!

PR #6, opened 2019-12-18
display and link to library notes

robertylewis:
Adds a page "Notes" that contains all notes declared with the `library_note` command, and links `Note [note name]` to the corresponding point on that page. There's only one instance of this in the library afaict, in the `algebra.category.Mon.basic` module doc.

Screenshot from 2019-12-18 17-54-20
Screenshot from 2019-12-18 18-08-47

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -227,7 +227,6 @@ def write_internal_nav(objs, filename, out):
 def write_notes_nav(notes, out):
   out.write('<h1>Lean <a href="https://leanprover-community.github.io">mathlib</a> docs</h1>')
   out.write('<h2><a href="#top">Library notes</a></h2>')
-  out.write('<div class="gh_link"></div>')
   for o in sorted([o[0] for o in notes]):
     out.write('<a href="#{0}">{0}</a><br>\n'.format(o))
 
@@ -341,6 +340,7 @@ badly formatted doc strings, or to add missing documentation.</p>
 """.format(mathlib_commit, mathlib_github_root, lean_root, lean_commit, search_snippet)
 
 notes_body = """
+<a id="top"></a>
 <h1>Lean mathlib notes</h1>
 
 <p>Various implementation details are noted in the mathlib source, and referenced later on.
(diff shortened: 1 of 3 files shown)


jcommelin:
> There's only one instance of this in the library afaict, in the `algebra.category.Mon.basic` module doc.

Only one instance of what? There are multiple notes, and multiple references to notes in mathlib already.

robertylewis:
Sorry, I could have been clearer there -- one *displayable* reference to a note. Most of the references are in line comments, e.g. [here](https://github.com/leanprover-community/mathlib/blob/f86abc7781d85564a9fba9f3deafe73453d22e8d/src/category_theory/limits/shapes/finite_products.lean#L26). These won't show up in the docs. I think the one I mention is the only one that appears in a declaration doc string or module doc.

gebner:
The note page contains lots of unrelated content---reducibility convention in the category theory library, instance priorities, details of the vector space definition, etc.  Would it make sense to display the notes instead in the files where there are defined in?  That is, e.g. `locally reducible category instances` in `algebra.category.Mon.basic`?

robertylewis:
I'm not sure. It makes sense for some of the notes, but not all of them. E.g. the instance priority notes are defined in `tactic.lint`. It's an unnatural place to link to, but there's no better place -- those notes feel kind of global.

robertylewis:
> I think the one I mention is the only one that appears in a declaration doc string or module doc.

Oh, there's actually another: one note references another note. After changing the linkification to accept lower case `note [...]`, this works just fine.

robertylewis:
We could categorize the notes, and use the categories to sort them on the notes page. The `library_note` command would take an extra argument:
```
library_note "lower instance priority" "Type class notes"
"Certain instances always apply during type-class resolution ..."
```
and then we'd have a section "Type class notes" in the docs.

It would probably look a little funny now, without many notes in the library, but might be better if the number of notes grows.

jcommelin:
I wouldn't bother. I would just view this note page as some sort of dump, where you only go if you clicked on some link...

PR #9, opened 2020-02-02
add mathjax to header so docs can use latex with the usual delimiters

glangmead:
This is one way to enhance the docs with LaTeX. The authors of mathlib can simply add math in the comments with the usual delimiters. These pass through into json_export.txt (escaped appropriately for json), and then directly into the HTML. I confirmed this end to end. The only change I had to make was to add MathJax to the header of each page, so it can find the delimited math and render it.

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -260,6 +260,12 @@ def write_body_content(objs, loc_map, filename, mod_docs, instances, body_out):
     else:
       write_mod_doc(o, loc_map, body_out)
 
+# MathJax configuration docs:
+#   http://docs.mathjax.org/en/latest/options/document.html#the-configuration-block
+# MathJax inlineMath/displayMath docs:
+#   http://docs.mathjax.org/en/latest/options/input/tex.html#tex-extension-options
+# StackOverflow link on why to avoid \(..\) and \[..\] when markdown is in the loop:
+#   https://math.meta.stackexchange.com/a/18714
 def html_head(title):
   return """<!DOCTYPE html>
 <html lang="en">
@@ -272,11 +278,20 @@ def html_head(title):
         <script>
         MathJax = {{
           tex: {{
-            inlineMath: [['$', '$'], ['\\(', '\\)']]
+            inlineMath: [['$', '$']],
+            displayMath: [['$$', '$$']]
           }},
-          svg: {{
-            fontCache: 'global'
-          }}
+          options: {{
+              skipHtmlTags: [
+                  'script', 'noscript', 'style', 'textarea', 'pre',
+                  'code', 'annotation', 'annotation-xml',
+                  'decl', 'decl_meta', 'attributes', 'decl_args', 'decl_header', 'decl_name',
+                  'decl_type', 'equation', 'equations', 'structure_field', 'structure_fields',
+                  'constructor', 'constructors', 'instances'
+              ],
+              ignoreHtmlClass: 'tex2jax_ignore',
+              processHtmlClass: 'tex2jax_process',
+          }},
         }};
         </script>
         <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
--- a/print_docs.py
+++ b/print_docs.py
@@ -268,6 +268,18 @@ def html_head(title):
         <link rel="shortcut icon" href="https://leanprover-community.github.io/assets/img/lean.ico">
         <title>mathlib docs: {1}</title>
         <meta charset="UTF-8">
+        <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
+        <script>
+        MathJax = {{
+          tex: {{
+            inlineMath: [['$', '$'], ['\\(', '\\)']]
+          }},
+          svg: {{
+            fontCache: 'global'
+          }}
+        }};
+        </script>
+        <script id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
     </head>
     <body>
         <div class="row">""".format(site_root, title)


print_docs.py
```diff
@@ -268,6 +268,18 @@ def html_head(title):
         <link rel="shortcut icon" href="https://leanprover-community.github.io/assets/img/lean.ico">
         <title>mathlib docs: {1}</title>
         <meta charset="UTF-8">
+        <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
+        <script>
+        MathJax = {{
+          tex: {{
+            inlineMath: [['$', '$'], ['\\(', '\\)']]
```
robertylewis: IIRC Zulip (and other LaTeX/makdown combos?) don't use single `$` delimiters for inline MathJax. It results in a lot of false positives. I'm not sure how big a worry it is here. Is MathJax smart enough to avoid inserting itself in code blocks `f $ g $ h a`? If not, double `$$` is safer.

print_docs.py
```diff
@@ -268,6 +268,18 @@ def html_head(title):
         <link rel="shortcut icon" href="https://leanprover-community.github.io/assets/img/lean.ico">
         <title>mathlib docs: {1}</title>
         <meta charset="UTF-8">
+        <script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
```
jcommelin: What is this `polyfill` thing? Is there a reason to grab mathjax from there? Should we self-host a copy?

print_docs.py
jcommelin: Is there a way to see how the docs render with this patch in place? I.e., can we test what happens?

robertylewis: http://robertylewis.com/mathlib_docs/ is a build with @glangmead 's changes enabled.

robertylewis: Something funny is going on [here](https://robertylewis.com/mathlib_docs/conv.html). I don't think the parens are escaped in the markdown. Does ['\\(', '\\)'] mean that everything in parens will become Mathjax??

jcommelin: It *should* mean that `\( x + \sqrt{y + z} \)` is parsed as mathjax. So the `\` has to be explicitly in the source before the parens.

robertylewis: I mean, that's what I assumed, but not what's happening.

glangmead: I see the problem with conv. I'll look into that, we can't have false positives. I note that conv starts life as a markdown doc, maybe there's something different about that code path. I'll try to break a .lean file in the same way to check.

print_docs.py
glangmead: The MathJax docs said to put both these inclusions which surprised me. Polyfill is apparently there as a backoff for "older browsers". I don't know how necessary it really is, I'll find out. As for self-hosting I don't know all the tradeoffs. The authors of mathlib don't need to care of course.

print_docs.py
robertylewis: It seems to be happening with everything that's surrounded by parens: https://robertylewis.com/mathlib_docs/geometry/manifold/manifold.html

glangmead: Agreed. Probably the parens in my configuration array need more backslashes. On the other hand if we're removing $ then I think we can accept the defaults and I can remove that line. I'll try both so that I understand what's going on.

Note that double-dollar and backslash-bracket will switch to display mode, which may not be what zulip is doing. That makes backslash-paren the only way to do inline math from a lean file. I'll see if MathJax has a way to exclude its processing from regions of the doc so we could have the best of both worlds and support dollars in the comments.

glangmead: Gosh is there an existing bug with parens? See https://leanprover-community.github.io/mathlib_docs/geometry/manifold/real_instances.html where the definition of euclidean_space in lean is `def euclidean_space (n : ℕ) : Type := (fin n → ℝ)` but on the web some of it is deleted: `def euclidean_space : ℕ → Type`.

glangmead: MathJax does exclude itself from `<code>` blocks and you can configure it to exclude more css classes. I'll test that out.

robertylewis: For what it's worth, MathJax is already smart enough to avoid running in `<pre>` or `<code>` blocks. So `$` occurrences in Lean code shouldn't trigger it, unless someone has forgotten to put backticks around it. So maybe the simple solution is to keep single `$` and remove `\(` as delimiters. I doubt there are many if any plain text lines in mathlib that have two `$`s.

robertylewis: > Gosh is there an existing bug with parens? See https://leanprover-community.github.io/mathlib_docs/geometry/manifold/real_instances.html where the definition of euclidean_space in lean is `def euclidean_space (n : ℕ) : Type := (fin n → ℝ)` but on the web some of it is deleted: `def euclidean_space : ℕ → Type`.

This isn't a bug. The web docs are only displaying the type, which is `ℕ → Type`. The `ℕ` argument doesn't have to be bound to print `Type`, but in the source it's referenced in the body of the definition `(fin n → ℝ)`.

robertylewis: It is maybe a badly phrased doc string though. It should "informally bind" `n`, i.e. "`euclidean_space n` is the space ℝ^n."

print_docs.py
glangmead: The docs http://docs.mathjax.org/en/latest/output/browser.html say that polypill is there to support IE11. The stats here https://www.w3counter.com/globalstats.php say that IE11 is ranked 8th with 2.3% of... whatever they're counting.

jcommelin: Ok, thanks for explaining. In that case, I'm fine with using polyfill.

glangmead: Thanks for the question. I only followed up with that research once you asked, having merrily copied and pasted it.

print_docs.py
glangmead: I learned a few things and will update the PR shortly. First, I've swung back to wanting dollar signs and NOT the parens or brackets, since the latter do not pass cleanly through markdown. Stackoverflow has the same problem, see https://math.meta.stackexchange.com/a/18714. So: only single and double dollars should be processed, which will be what most people guess they should use anyway. 

Further, I've added all the lean-related css classes Rob defined in the python script as exclusions for MathJax. Some of those do not also have pre or code around them, so this should help keep MathJax out of the Lean stuff entirely. It can be further tweaked later of course!

glangmead:
Attaching a zipped version of real_instances.html, where I have a test comment

```# Constructing examples of manifolds over dollar $\mathbb{R}$ or double dollar $$\mathbb{R}$$. Or paren \(\mathbb{R}\) or bracket \[\mathbb{R}\]```

and the rendered html attached here shows that dollars are processed by mathjax, double-dollars are processed and create display mode, and the paren/brackets are not processed by mathiax.

[real_instances.html.zip](https://github.com/leanprover-community/doc-gen/files/4148429/real_instances.html.zip)

Happy also for @robertylewis to regenerate the docs so we can browse for issues again if desired.

robertylewis:
I've regenerated the docs, with a few examples of good behavior: https://robertylewis.com/mathlib_docs/tactics.html 

I don't know of any natural occurrences of `$..$` in the library to check, but I'm pretty convinced that this will work! We'll keep an eye out for any breakage.

PatrickMassot:
For the record, let me write that I think we should do server-side rendering here. There is no dynamic content in those pages, the doc generation script should call mathjax instead of asking all readers web browsers to do so. This allows opens up the possibility of having commutative diagrams or other fancier LaTeX. But I don't have time right now :-(

glangmead:
When you say "render", what output format do you have in mind? To me LaTeX is a markup language like HTML and MathML that the browser's job is to render. Because it's left as markup the browser can change the font, colors, do screen reading for accessibility etc.

urkud:
@PatrickMassot AFAIK, MathJax outputs different HTML in different browsers, so it can only run in browser. At least it worked that way a couple of years ago.

PR #11, opened 2020-02-06
display tactic/command/etc entries using mathlib command

robertylewis:
https://github.com/leanprover-community/mathlib/pull/1963 replaces `tactics.md`, etc with a new command for adding doc information. This is the corresponding update on this side, extracting the information added in those commands:

Screenshot_20200206_135621

Still to do: JavaScript filtering, to display only entries that have certain tags.

This should NOT be merged until the above PR lands and we migrate the doc entries from the md file to the new command.

cc @bryangingechen


bryangingechen:
This looks great overall though I haven't tried to run it locally yet. One thing to check (probably in the associated mathlib PR) is that linking between tactic docs is easy to do. From a first glance it looks like the URLs will be fairly predictable (something like `tactic.html#decl_name`) which is good. Another thing to do is to add some info about this new feature to the README.

Would it be possible to avoid hard-coding mathlib-specific docs / doc commands into `print_docs.py` and `export_json.lean`? I think this project has the potential to become usable for Lean repos outside mathlib, and keeping "data" separated from the code as much as possible now will be easier than extricating it in a refactor later on.

robertylewis:
I think the things that are hardcoded in the tactic doc setup are:
* the entry categories, e.g. tactic, command, hole command
* import paths: we display either `import tactic.core` or `import tactic.default`, if either works, and the precise import where the tactic is defined

Do you see others?

There are other hardcoded things in the doc gen tool, mainly because it looks at paths to distinguish mathlib and core files, and it finds specific .md files in the mathlib repo. I think it would be easy enough to generalize this to core vs other repos.

robertylewis:
I've updated this so it builds with https://github.com/leanprover-community/mathlib/pull/2114 . There are definitely some rough edges still. It needs header text for the attributes page, and I'd like to implement a way to filter tactic/command/attribute entries by tag.

robertylewis:
FYI, this is what it looks like right now: https://robertylewis.com/mathlib_docs/tactics.html

bryangingechen:
Overall it looks nice! 

- I noticed that we're showing both the text above and below the horizontal rule in entries like this: https://robertylewis.com/mathlib_docs/tactics.html#rename'

- Also, lack of linebreaks before `---` led to weirdness here and a few other places; I've made some suggestions at the mathlib PR that I'll commit soon: https://robertylewis.com/mathlib_docs/tactics.html#suggest

- Do you know what's going on here? https://robertylewis.com/mathlib_docs/commands.html#new%20cmd

Two thoughts for future PRs:

- We should add syntax highlighting. It looks like we're already using the `fenced-code-blocks` extension for markdown2 and [it supports highlighting via pygments](https://github.com/trentm/python-markdown2/wiki/fenced-code-blocks). I guess we should revisit pygments anyways since it's still broken in Zulip.

- It might be nice if we could click on the import statements (e.g. `import tactic.XXX`) to copy them to the clipboard. (I could hack this in later.)

(I was a little confused because the links in the sidebar redirect to the docs at leanprover-community.github.io instead of robertylewis.com, but that's not really relevant.)

bryangingechen:
> Do you know what's going on here? https://robertylewis.com/mathlib_docs/commands.html#new%20cmd

The [`reassoc_axiom` doc entry](https://github.com/leanprover-community/mathlib/blob/56da7b378a2fa4c98360732f89069d6da72d7d7f/src/tactic/reassoc_axiom.lean#L167) is missing so maybe something strange happened there?

bryangingechen:
Also, `linarith` is showing up [in the commands](https://robertylewis.com/mathlib_docs/commands.html#linarith), not the tactics... maybe you're using a different commit than I expected?

robertylewis:
> * I noticed that we're showing both the text above and below the horizontal rule in entries like this: [robertylewis.com/mathlib_docs/tactics.html#rename](https://robertylewis.com/mathlib_docs/tactics.html#rename)'

Oh, yes, this is also on the to-do list.

> * Do you know what's going on here? [robertylewis.com/mathlib_docs/commands.html#new%20cmd](https://robertylewis.com/mathlib_docs/commands.html#new%20cmd)

This (and `linarith`) were junk code in the `doc_gen` repo I was using for testing. It's fixed -- maybe just locally for now, but nothing to worry about.

robertylewis:
Okay: this is updated to have a JS filter on the tactic/command/attribute pages. See https://robertylewis.com/mathlib_docs/tactics.html 

Any entry that doesn't have any tags gets the "untagged" tag added, to make them easy to identify. (Maybe the `add_tactic_doc` command should enforce that there's at least one tag?)

robertylewis:
@bryangingechen @gebner have a look; this basically finishes my to-do list. One more thing: sometimes entries are repeated in the "import using" menu, I'll fix that right now.

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+
+
+function filterSelectionClass(c, classname) {
+    var x, i;
+    x = document.getElementsByClassName(classname);
+    //if (c == "all") c = "";
+    //cs = c.split(",");
+    // Add the "show" class (display:block) to the filtered elements, and remove the "show" class from the elements that are not selected
+    for (i = 0; i < x.length; i++) {
+      w3AddClass(x[i], "hide");
```
gebner:
```suggestion
      x[i].classList.add("hide");
```

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+      for (j = 0; j < c.length; j++) { 
+          if (x[i].className.indexOf(c[j]) > -1) w3RemoveClass(x[i], "hide");
```
gebner:
```suggestion
          if (x[i].classList.contains(c[j])) {
            x[i].classList.remove("hide");
          }
```

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+
+
+
+function filterSelectionClass(c, classname) {
+    var x, i;
+    x = document.getElementsByClassName(classname);
+    //if (c == "all") c = "";
+    //cs = c.split(",");
+    // Add the "show" class (display:block) to the filtered elements, and remove the "show" class from the elements that are not selected
+    for (i = 0; i < x.length; i++) {
```
gebner:
```suggestion
    for (const elem of document.getElementsByClassName(classname)) {
```

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+    }
+  }
+
+  function filterSelection(c) {
+    filterSelectionClass(c, "tactic");
+    filterSelectionClass(c, "taclink");
+  }
+
+// Show filtered elements
+function w3AddClass(element, name) {
```
gebner:
```suggestion
```

`element.classList.{add,remove}` is now widely supported and we should use it: https://caniuse.com/#feat=classlist

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+    //alert(getSelectValues(select));
+    filterSelection(getSelectValues(select));
+}
+
+function getSelectValues(select) {
+    var result = [];
+    var options = select && select.options;
+    var opt;
+  
+    for (var i=0, iLen=options.length; i<iLen; i++) {
```
gebner:
```suggestion
    for (const opt of options) {
```

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
...
+
+function filterSelectionClass(c, classname) {
+    var x, i;
+    x = document.getElementsByClassName(classname);
+    //if (c == "all") c = "";
+    //cs = c.split(",");
+    // Add the "show" class (display:block) to the filtered elements, and remove the "show" class from the elements that are not selected
+    for (i = 0; i < x.length; i++) {
+      w3AddClass(x[i], "hide");
+      for (j = 0; j < c.length; j++) {
```
gebner:
```suggestion
      for (const tagName of c) { 
```

nav.js
```diff
@@ -63,4 +63,83 @@ for (const impl_collapsed of document.getElementsByClassName('impl_collapsed'))
         impl_args[0].addEventListener('click', () =>
             impl_collapsed.classList.remove('impl_collapsed'));
     }
-}
\ No newline at end of file
+}
+
+
+
+function filterSelectionClass(c, classname) {
```
gebner:
```suggestion
function filterSelectionClass(tagNames, classname) {
```

I hope that's what's in `c`.

gebner:
I think the filtering functionality (i.e. filtering the tactics) is useful, but the drop-down is neither easy nor comfortable to use.

At first I tried to click on some tags, but that only removed the tactics with that tag (which didn't help me).  I think it would be more useful if: nothing selected = all tactics are shown.  And: one tag selected = tactics with that tag are shown.

Then I figured out that you can click "select all" to get a blank list.  I had to press twice since I already had something selected.

We don't have that many tags.  I would prefer to show all of them at once.  Then I wouldn't have to scroll.  If I already know the tag, then I probably know the name of the tactic as well and can just search for the tactic itself.

nav.js
robertylewis: Ah, thanks! My JS knowledge doesn't go beyond copying code from StackOverflow.

robertylewis:
Thanks, that's helpful feedback -- I'll see what I can do!

bryangingechen:
This looks pretty good to me. The list of tags under each doc entry would be more useful if they were clickable. Also, I'm not sure they need to be hidden under a `<summary>` tag. Maybe they can always be displayed with a smaller font / less spacing.

I think `import tactic.default` should be transformed to `import tactic`.

Also, I think the `import with` display would be most useful if it always showed the most precise import and then either `import tactic.basic` or `import tactic` since those are the two main collections of imports we provide. Not sure what things look like after your changes, but currently it doesn't seem like `tactic.basic` shows up anywhere in those sections.

robertylewis:
I think I made all of Gabriel's suggestions. The demo is updated. Will work on Bryan's in a bit!

robertylewis:
I've made most of Bryan's suggestions. The tags are no longer in a summary, and we display `tactic.basic` as an import instead of `tactic.core`.

I'm not sure about making the tags clickable. Should clicking on a tag filter the display to only show entries with that tag? (And update the checks in the drop down menu to match?) I could try this -- it's slightly more effort. An alternative would be to make the tags look less like buttons.

bryangingechen:
> Should clicking on a tag filter the display to only show entries with that tag? (And update the checks in the drop down menu to match?)

Yeah, that's what I was thinking. We could add it later if it's too much trouble now.

robertylewis:
Yeah, let's wait on that. I'd like to get this merged before we submit the library maintenance paper.

bryangingechen:
OK, I'm pretty happy with this as is. After this is merged I guess the next step is to do some cleanup on mathlib, e.g. deleting the old markdown files and fixing tags.

gebner:
Ok, I think it's much better now.

How attached are you to the multiselect javascript library?  I do not particularly like the dropdown: it is inconsistent with the `<details>` element we use elsewhere, and I don't even know where it comes from.  I also don't really see the benefits of hiding the tags: the list is short and we want users to see them and click them.  I'd prefer to have standard html checkboxes instead of the multiselect library, and if we really want to hide the tags, put them inside a `<details>` element.

If you'd rather merge the PR before the CICM submission, then I'm happy to change this later as well.

robertylewis:
> How attached are you to the multiselect javascript library?

Not particularly. It's a random script I found as an answer to a SO post, I think. 

> I also don't really see the benefits of hiding the tags:

Do you mean you'd rather display all the tags at the top of the left column permanently, instead of using a menu? There aren't a ton of them, but there are enough (especially on the tactics page) that I think it would look very cluttered, especially on a smaller screen.

I can try out putting standard checkboxes in a `<details>` element instead of the multiselect thing, that sounds reasonable.

gebner:
> There aren't a ton of them, but there are enough (especially on the tactics page) that I think it would look very cluttered, especially on a smaller screen.

That's probably true, I've been testing it on a 32 inch monitor.

> I can try out putting standard checkboxes in a <details> element instead of the multiselect thing, that sounds reasonable.

I'd prefer this solution.  I've also just tried the docs on my phone and the multiselect thing causes the keyboard to pop up, which is really annoying.

robertylewis:
Okay, I like putting the checkboxes in a `<details>` element!

PR #13, opened 2020-02-08
Markdown avoid math delimiters

glangmead:


glangmead:
@robertylewis This PR is inadequate in that misaka does not support "cuddled lists" which are used e.g. here https://leanprover-community.github.io/mathlib_docs/algebra/category/Mon/basic.html. But I'm out of ideas for now.

robertylewis:
Hmm. I suppose we could open an issue with the Misaka team (https://github.com/FSX/misaka/issues) asking if there's any possibility of supporting cuddled lists. Or alternatively with markdown2 (https://github.com/trentm/python-markdown2) asking if there's any possibility of math mode/arbitrary escape sequence delimiters.

If neither of those are likely, I guess switching to Misaka and replacing cuddled lists is the better option. Math mode will never be reliable with the current markdown2 setup.

glangmead:
Closing this so I can redo it with the avoidance of cuddled lists in mind. I lost my train of thought over the intervening weeks!

Issue #10, opened 2020-02-04
Bad MathJax/markdown interactions


robertylewis:
Putting the line `what if I put $f[a,b](cd)$ in tex? $g_{x}(y)$ ?` in a markdown file/block produces:
![Screenshot_20200204_131050](https://user-images.githubusercontent.com/4967469/73743874-d7660f80-474f-11ea-8690-b423c68a25dc.png)

I'm not sure if/how I can disable markdown processing between delimiters using markdown2. 

cc @glangmead

gebner:
Oh, `f[a,b](cd)` is a Markdown link...  I guess this is why gitlab uses ``$`a^2+b^2=c^2`$``: https://docs.gitlab.com/ee/user/markdown.html#math

glangmead:
I think I see. We want markdown2 to be hands-off between the math delimiters just like we want mathjax to be hands-off outside them. I see two solutions: find a markdown library that supports this, or submit a patch to markdown2 to add another "extra" that does this. I've spent a few minutes looking into the latter and can only say that it seems doable. In spirit it should be similar to its logic for handling backticks, since it's also hands-off between those.

glangmead:
See `_do_code_spans` at https://github.com/trentm/python-markdown2/blob/master/lib/markdown2.py

glangmead:
Here's a python markdown library that appears to support ignoring what's between math delimiters: https://github.com/FSX/misaka

jcommelin:
Doesn't `pandoc` do the right thing as well? https://pandoc.org

glangmead:
I had success with your example @robertylewis by changing print_docs.py to use this:

```pip install misaka``

then instantiate the global markdown converter object:

```
import misaka
markdown_to_html = misaka.Markdown(misaka.HtmlRenderer(), extensions=('math', 'math-explicit',))
```

and use it where you call markdown2 now. You configured markdown2 to do some other things with its extras, namely 'doc' and ['code-friendly', 'cuddled-lists', 'fenced-code-blocks', 'link-patterns']. I haven't looked up how to do those with misaka. Internally it's using something called hoedown which is another place to learn how it works.

https://github.com/hoedown/hoedown/

glangmead:
Oh AND, misaka converts the dollar signs into `\(..\)` and double-dollars into '\[..\]' so we need to reintroduce those as allowable for mathiax.

In the python that means quadruple backslashes:

```python
          tex: {{
[...]

PR #15, opened 2020-03-31
https://github.com/leanprover-community/doc-gen/issues/10

glangmead:
- switch to misaka for converting markdown to HTML in most places
- keep markdown2 for generating TOCs
- misaka doesn't support cuddled lists which appear in mathlib, so a mathlib patch will remove those

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -11,6 +11,7 @@ import os
 import glob
 import textwrap
 import markdown2
+import misaka
 import re
 import subprocess
 import toml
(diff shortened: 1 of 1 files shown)


glangmead:
I'm still looking over the consequences of switching to misaka. It has some differences I'm not sure I like, such as how treple-backticks behave within lists.

glangmead:
Misaka introduces an issue of its own which is that it fails to recognize code blocks that are not separated from the text by blank lines. Instead it treats them as inline code which is not what anyone expects. So my ad-hoc approach here is not going to pan out. The right approach is to decide what flavor of markdown we are adopting and only use converters that adhere to that flavor, and within that constraint address bugs with math.

PR #19, opened 2020-05-15
update links to tutorials

bryangingechen:
As promised in leanprover-community/mathlib#2676, this PR updates the links to the tutorials that have been moved to the community website.


robertylewis:
Thanks! Do you think it's better to keep the tutorials linked from the docs, or to remove the links entirely? I'm not sure. They're not really "part of the docs," it's more that this was the only convenient place to host them before the new website.

bryangingechen:
Hmm, I'm not sure. I originally removed the links entirely and just added a reference to them to the index page, but then I wondered whether someone that had browsed the docs before and noticed the tutorial links there would be confused that they disappeared.

PR #20, opened 2020-05-15
Style: make <, ≤, >, ≥ looks better by removing links

RaitoBezarius:
Those are only enabled upon hovering on them.

### DIFF
--- a/style_js_frame.css
+++ b/style_js_frame.css
@@ -227,14 +227,20 @@ code {
 code a:link, code a:visited, code a:active,
 .decl a:link, .decl a:visited, .decl a:active {
     color: black;
-    /** Don't show underline on types, to prevent the ≤ vs < confusion. **/
+}
+
+/** Don't show underline on types, to prevent the ≤ vs < confusion. **/
+.decl_type a:link, .structure_field a:link {
     text-decoration: none;
+}
+
+/** Show it on hover though. **/
+.decl_type a:hover, .structure_field a:hover {
+    text-decoration: underline;
 }
 
 code a:hover, .decl a:hover {
     color: darkslategray;
-    /** Show it on hover though. **/
-    text-decoration: underline;
 }
 
 .indent {
--- a/style_js_frame.css
+++ b/style_js_frame.css
@@ -227,10 +227,14 @@ code {
 code a:link, code a:visited, code a:active,
 .decl a:link, .decl a:visited, .decl a:active {
     color: black;
+    /** Don't show underline on types, to prevent the ≤ vs < confusion. **/
+    text-decoration: none;
 }
 
 code a:hover, .decl a:hover {
     color: darkslategray;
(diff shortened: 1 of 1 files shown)


bryangingechen:
Here's a screenshot of what things look like after your change:

Screenshot 2020-05-15 21 49 46

One thing I don't like is that in `hx : is_integral α x`, I can't tell that `is_integral` is a ilnk and that `hx`, `α` and `x` are not. Could you tweak the colors or some other feature to make the links stand out somehow?

RaitoBezarius:
Guessed that it could create some issues, here's another patch which
makes things a bit different, it still gives us links in hypothesis AFAIK.

As order stuff has rarely hypothesis but most stuff seems to be in
implications, etc.

Let me know how it goes. :)

--  
Ryan

RaitoBezarius:
Screen Capture_select-area_20200516040210
Here's some dark theme screenshot (sorry for the dark theme! :D)

bryangingechen:
Hmm, I actually kind of liked it more with all underlines removed by default, since they overlap with the underscores. Could you somehow make it so the non-links are more grayed out or less bold or something? 

In your dark theme, I like the fact that for the second declaration `theorem minimal_polynomial.monic`, `hx`, `α` and `x` are actually a slightly different color which distinguishes them from the links. Not sure why that doesn't happen for `def minimal_polynomial` though. 

Sorry for being picky!

RaitoBezarius:
> Hmm, I actually kind of liked it more with all underlines removed by default, since they overlap with the underscores. Could you somehow make it so the non-links are more grayed out or less bold or something?
We can definitely do something like this, I'll do it tomorrow, if that's okay.
> 
> In your dark theme, I like the fact that for the second declaration `theorem minimal_polynomial.monic`, `hx`, `α` and `x` are actually a slightly different color which distinguishes them from the links. Not sure why that doesn't happen for `def minimal_polynomial` though.
That's actually a side effect of a CSS filter from my GPU… :'D (I use the Dark Reader extension with Filter+).

> 
> Sorry for being picky!
Don't worry, it's all about the *user experience*.

rwbarton:
In my opinion, even just deleting `color: black` and adding `text-decoration: none` looks fine, and definitely an improvement on the status quo:
image
I don't think the underlines are really needed on hover either, since the text color already changes on hover (at least in my browser).

rwbarton:
Now I realize that links in the `decl_args` part (between the theorem name and the colon) aren't visible with my change, because their color gets overridden. So I would suggest taking the original commit, but also changing `color: black;` to `color: #0000ee` (this seems to be the usual default link color: https://stackoverflow.com/questions/4774022/whats-default-html-css-link-color)

gebner:
Please make follow-up PRs if you like something else.  But if we don't merge something now, then this will never be fixed.

PR #21, opened 2020-05-19
add page to look up by decl name

robertylewis:
This adds a page `find.html` to the docs. Going to `find.html?decl=nat.add` will redirect you to `core/init/core.html#nat.add`.

Usual disclaimers apply, I don't know anything about web development, probably passing a big json object as a string in a js file isn't the optimal way to do this. What's the better way?

### DIFF
--- a/find.html
+++ b/find.html
@@ -1,0 +1,6 @@
+<html>
+    <head>
+        <script src="json_output.js"></script>
+        <script src="redirect.js"></script>
+    </head>
+</html>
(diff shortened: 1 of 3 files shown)


gebner:
I think I told you about my preferred solution on Sunday: create one file per declaration, which contains just a redirect.  For example, create `declaration/nat.add.html` with the following content:
```html
<meta http-equiv="refresh" content="0;url=../core/init/core.html#nat.add">
```

This requires no javascript, and IMHO the url looks much nicer.  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

robertylewis:
#22 seems better; we'll have less control over error messages, but otherwise it seems cleaner. Closing this.

PR #22, opened 2020-05-20
new attempt at declaration lookup

robertylewis:
A new attempt at #21 based on @gebner 's suggestion. I was a little skeptical at first, but this does indeed seem to be better. See it in action: 
https://robertylewis.com/mathlib_docs/find/manifold

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -49,7 +49,7 @@ html_root = root + '/' + (cl_args.t if cl_args.t else 'html/')
 site_root = "http://localhost:8000/"
 
 # web root, used in place of `site_root` if the `-w` flag is used
-web_root = "https://robertylewis.com/mathlib_docs/"
+web_root = "https://leanprover-community.github.io/mathlib_docs/"
 
 # root directory of mathlib.
 local_lean_root = root + '/' + (cl_args.r if cl_args.r else '_target/deps/mathlib/')
--- a/print_docs.py
+++ b/print_docs.py
@@ -646,6 +646,20 @@ def write_site_map(partition):
     out.write(site_root + filename + '.html\n')
   out.close()
 
+def redirect_body(decl_name, decl_loc):
+  url = filename_core(site_root, decl_loc, 'html')
+  return f"""
+<!DOCTYPE html>
+<Content-Type="text/html">
+<meta http-equiv="refresh" content="0;url={url}#{decl_name}">
+"""
+
+def write_redirects(loc_map):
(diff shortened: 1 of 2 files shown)


robertylewis:
This doesn't work with the suggested local testing using `python -m http.server` since it won't serve files without .html extensions with the right headers. GitHub Pages does this though.

print_docs.py
```diff
@@ -646,6 +646,20 @@ def write_site_map(partition):
     out.write(site_root + filename + '.html\n')
   out.close()
 
+def redirect_body(decl_name, decl_loc):
+  url = filename_core(site_root, decl_loc, 'html')
+  return f"""
+<!DOCTYPE html>
+<Content-Type="text/html">
```
gebner, line 653:
```suggestion
<meta http-equiv="content-type" content="text/html">
```
Are these two lines necessary at all?

robertylewis: Probably not, I'll test. This was part of my failed attempt to make it work with the Python local server.

gebner:
An idea to make `http.server` work: we could generate files like `find/nat.add/index.html` and then link to `find/nat.add`.

robertylewis:
Ah, much better!
https://robertylewis.com/mathlib_docs/find/units.coe_dvd

robertylewis:
While we're at it, maybe it's time to make CI force push to the `mathlib_docs` repo. There's no need to track history there. And this will add > 40k files to that repo. They won't change all that much, but still.

robertylewis:
A new option: we can also redirect to location in the source code.
https://robertylewis.com/mathlib_docs/find/matrix.det/src

robertylewis:
I'm done adding features; ready when @gebner and/or @bryangingechen are happy!

print_docs.py
```diff
@@ -99,6 +99,27 @@ def library_link(filename, line=None):
...
+  return False
+    
+
+def library_link_from_decl_name(decl_name, decl_loc, file_map):
+  try:
+    e = next(d for d in file_map[decl_loc] if name_in_decl(decl_name, d))
+  except StopIteration:
+    if decl_name[-3:] == '.mk':
+      return library_link_from_decl_name(decl_name[:-3], decl_loc, file_map)
+    print(decl_name)
```
gebner, line 118: This is debugging code, right?

robertylewis: Indeed. I've left it in, but clarified the debug output.

PR #26, opened 2020-06-03
style: make links blue; remove underlines consistently

rwbarton:


rwbarton:
image
Not sure how to get it in the screenshot, but links have underlines (and change color) when hovered.

PR #27, opened 2020-06-04
Github Actions

bryangingechen:
This workflow builds mathlib docs for each commit pushed to any branch and additionally deploys the built docs to the web for each commit to master. (This way we don't have to manually trigger a build on mathlib to update the docs.)


gebner:
I think it is a good idea to have CI here as well.  I'm not so sure I like the duplication of the deployment logic and responsibility.  I'm pretty sure we'll forget to update one of them in the future.

bryangingechen:
Good point. As far as reducing deployment logic, I can try to make one script (merging `ci.sh` here and `deploy_docs.sh` in mathlib) that we can run both here and in mathlib's workflow. Would it be better to host that here or in mathlib?

bryangingechen:
OK, this now uses the `deploy_docs.sh` script from mathlib.

PR #45, opened 2020-08-12
fix linkification in code blocks

robertylewis:
This linkifies all identifiers found in code blocks, instead of just the ones lucky enough to be caught before.

Then:
image

Now:
image

### DIFF
--- a/leanpkg.toml
+++ b/leanpkg.toml
@@ -1,8 +1,8 @@
 [package]
 name = "."
 version = "0.1"
-lean_version = "leanprover-community/lean:3.17.1"
+lean_version = "leanprover-community/lean:3.18.4"
 path = "src"
 
 [dependencies]
-mathlib = {git = "https://github.com/leanprover-community/mathlib", rev = "f78a012d976ebf3f2a4694fa16cff1632ce0b57f"}
+mathlib = {git = "https://github.com/leanprover-community/mathlib", rev = "4324778f62a42d4d0cba64680f2362f41b4a8a73"}
--- a/print_docs.py
+++ b/print_docs.py
@@ -201,7 +201,7 @@ def linkify(string, file_map):
 def linkify_type(string, loc_map):
   splitstr = re.split(r'([\s\[\]\(\)\{\}])', string)
   tks = map(lambda s: linkify(s, loc_map), splitstr)
-  return "".join(tks)
+  return f'<code>{"".join(tks)}</code>'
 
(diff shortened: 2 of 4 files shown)


robertylewis:
This has the side effect of linkifying things like `ring` [here](https://leanprover-community.github.io/mathlib_docs/tactics.html#ring). Again, it was an accident that this wasn't happening before, since there's no way for the tool to distinguish an abbreviated reference to `tactic.interactive.ring` from a full one to `ring`. I'm not sure there's anything that can be done here.

bryangingechen:
The fact that we have both a tactic and a type named `ring` also causes some weird error messages when users forget to `import tactic`.

robertylewis:
I'm not sure what the line
```python
  env.filters['linkify_type'] = lambda x: linkify_type(x, loc_map)
```
does in `setup_jinja_globals` or if the change to `linkify_type` affects this. @gebner ?

kbuzzard:
@bryangingechen I saw a variant of this the other day -- you know we have a `group` tactic right? ;-)

bryangingechen:
> @bryangingechen I saw a variant of this the other day -- you know we now have a `group` tactic right? ;-)

Yeah, although we went with `field_simps` instead of `field`. I can imagine someday we might have a `def` named `exact` too.

I'm not sure whether we should do anything about this here. Maybe it's worth discussing on Zulip?

gebner:
> This has the side effect of linkifying things like ring here.

I'm pretty sure I've seen this already with other tactics before.  :shrug: 

> I'm not sure what the line `env.filters['linkify_type'] = [...]` does.

This registers a *filter* in jinja: once you've added a function in this way, you can call it from a template with the `{{ ... | linkify_type }}` syntax.

gebner:
The `linkify_type` was the right choice all along for the filter.  We only used it to add links to declaration names, for which `linkify` is just as good.

https://github.com/leanprover-community/doc-gen/pull/46  This also adds links inside `` ``` `` code blocks.

robertylewis:
Hmm, something's wrong with the code block linkification. If you look at `tactics.html#case` you see `nat.zero` getting linkified wrong:
image
It handles inline `nat.zero` correctly.

gebner:
> Hmm, something's wrong with the code block linkification.

This works just fine here.  What pygments version are you using?

hier_name_hl

robertylewis:
Huh. Apparently the wrong one, it looks fine on the live docs too.

PR #47, opened 2020-08-13
Automatic expression formatting using CSS.

gebner:
Thanks to @DanielFabian, who introduced this technique in https://github.com/leanprover-community/mathlib/pull/3764

This probably requires a bit more testing, but it's too late for that today.

Demo: https://gebner.github.io/mathlib_docs/logic/relator.html#relator.rel_forall_of_total (resize the browser horizontally)


DanielFabian:
that's just plain sexy ;-)

gebner:
Let's test this in production.

PR #48, opened 2020-08-14
Print opt_param and auto_param using syntax.

gebner:
The auto_param printing is a bit wonky because the pretty printer cannot produce name literals:
```lean
∀ α, ... . "control_laws_tac"
```

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -345,3 +345,7 @@ do handle ← mk_file_handle filename mode.write,
 meta def main : io unit :=
 export_json "json_export.txt"
 
+-- HACK: print gadgets with less fluff
+notation x ` := `:10 y := opt_param x y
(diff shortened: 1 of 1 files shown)


gebner:
@robertylewis @bryangingechen Any opinions on this?  Some declarations are pretty much unreadable at the moment due to auto params: https://leanprover-community.github.io/mathlib_docs/geometry/manifold/smooth_manifold_with_corners.html#model_with_corners.mk_coe

robertylewis:
Oh, I completely missed this PR, sorry. It looks right in the output? If so I'm good with merging as-is.

gebner:
It's not perfect because it prints `continuous f . continuity'` as `(continuous f . "continuity'")`, but at least it's better than what we have now.

PR #52, opened 2020-08-25
feat(typeclass_stats): scrape typeclass hierarchy into JSON

jcommelin:


src/typeclass_stats.lean
```diff
@@ -0,0 +1,122 @@
...
+structure node extends item.
+
+namespace node
+
+def jsonify : node → string :=
+λ n,
+"{ \"name\" : \"" ++ n.name ++ "\",\n" ++
+  "\"topic\" : \"" ++ n.topic ++ "\",\n" ++
+  "\"file\" : \"" ++ n.file ++ "\",\n" ++
+  "\"line\" : " ++ n.line ++ " }\n"
```
eric-wieser, line 40: There's now an API for building json, right?

jcommelin: Yes, that's right. Feel free to hack on this. I mostly put up this PR so that we don't loose the script.

PR #53, opened 2020-08-26
Scroll current file to center.

gebner:
At the moment, the javascript scrolls the current file in the left
navbar to the top of the screen, which I find a bit irritating because
then you can't see the files above it anymore.  This PR changes the
behavior so that it is scrolled into the vertical center of the screen.


gebner:
I've tested it locally and it seems to work without the `setTimeout` (with the timeout the scrolling happens much later, and this is irritating).

PR #54, opened 2020-08-26
Gray-out implicit arguments instead of collapsing them

gebner:
At the moment, implicit arguments (including instance-implicit
arguments) are collapsed to `{…}`.  This approach impacts not only
discoverability but it also obscures the theorem statement by omitting
the type-class requirements.

It is particularly annoying that the type-class arguments are not shown
by default, because these are typically critical to understand the
theorem statement.

As an easy fix, this PR proposes to *gray-out* the implicit arguments
instead of programatically hiding them behind an unintuitive `{…}`
button.  As soon as the user hovers over the arguments, the arguments
are shown in full color.  This is done in an eye-pleasing way by setting
appropriate easing animations.

gray_out_example

### DIFF
--- a/nav.js
+++ b/nav.js
@@ -34,24 +34,6 @@ for (const currentFileLink of document.getElementsByClassName('visible')) {
 
 
 
-// Expansion of implicit arguments ({...})
-// ---------------------------------------
-
-
-for (const impl_collapsed of document.getElementsByClassName('impl_collapsed')) {
-    const impl_args = impl_collapsed.getElementsByClassName('impl_arg');
-    if (impl_args.length > 0) {
-        impl_args[0].addEventListener('click', () =>
-            impl_collapsed.classList.remove('impl_collapsed'));
-    }
-}
-
-
-
-
-
-
-
 // Tactic list tag filter
 // ----------------------
 
--- a/style.css
+++ b/style.css
@@ -438,19 +438,14 @@ a:hover {
     text-decoration: underline;
 }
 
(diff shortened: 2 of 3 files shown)


robertylewis:
I agree that hiding type class arguments can be confusing. On the other hand, there was a reason we collapsed implicit arguments to begin with: seeing long statements with lots of implicit arguments can be confusing itself. I'm not sure what the optimal solution is.

gebner:
In my limited testing, I found that graying out the implicit arguments was enough so that I could ignore them.  The contrast to the explicit arguments is large enough that at first glance you only see the explicit arguments.  The most significant difference is that there is now extra space between the explicit arguments (occupied by the grayed-out implicits).

An alternative would be to change the algorithm that determines which arguments are explicit.  I believe we should show all arguments by default, except for:
 * arguments whose type is a sort,
 * type-class arguments that occur in a following type-class argument as a dependency.

That is, `{α : Type u} [topological_space α] [ring α] [topological_ring α] {x : α} : x = x` should be rendered as `[topological_ring α] {x : α} : x = x` by default.
In addition all pis that have a binder name (different from `a`) should be shown as arguments.

robertylewis:
The downside to changing the algorithm, I guess, is that we still need a way to show that some arguments are hidden/to expand them. As you point out, the clickable `{...}` isn't exactly intuitive.

Let me build the docs with this PR when I have a chance to see how intuitive it feels. Might also be worth asking this on Zulip.

gebner:
Merging based on positive feedback on Zulip.

PR #55, opened 2020-08-26
Fix trailing parens on separate lines.

gebner:
This PR significantly simplifies the generation of nesting tags.  We
used to generate these from the group and nest constructors of the
`format` object and then apply various ad-hoc postprocessing operations.
The proposed approach instead extracts the nesting tags from the
expression structure that is reflected in the `tagged_format.tag`
constructors.  Only a single postprocessing step is required then:
we push adjacent parens into the nesting block so that they don't appear
on separate lines.  (We also coalesce adjacent strings for efficiency.)

I've also updated the JSON schema in the docstring to correctly describe
the currently generated JSON.

A side-effect of this construction is that the nesting tags now
correspond to the expression structure.  We could add a highlighting
setup similar to the one in the vscode info view widget.  However
so far I haven't been able to find a way to express this in CSS
in a way that is both helpful and not distracting.

Before:
trailing_parens_before

After:
trailing_parens_after


robertylewis:
It's like you're reading my mind. I saw an instance of this right before I left the office and made a note to look into it; this email was waiting for me when I got home!

PR #56, opened 2020-08-26
Show named arrows as arguments.

gebner:
Before:
more_arguments_before

After:
more_arguments_after

Particularly for meta declarations it is often an improvement:
more_arguments_improvement

It's a bit less useful than I'd thought though.  For once, we cannot distinguish explicitly named `a` arguments from arrows (at least before https://github.com/leanprover-community/lean/pull/442).  Also many recursive definitions sadly do not name their arguments.  For theorems it might be nice to have each hypothesis on a separate line.

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -146,23 +146,19 @@ meta def binder_info.is_inst_implicit : binder_info → bool
 | binder_info.inst_implicit := tt
 | _ := ff
 
-meta def count_named_intros : expr → tactic ℕ
-| e@(expr.pi _ bi _ _) :=
-  do ([_], b) ← open_n_pis e 1,
-     v ← count_named_intros b,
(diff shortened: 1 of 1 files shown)


robertylewis:
I'm sorry for sitting on this for so long -- I think this should be merged! I just played around with it a little and didn't see anything that looked bad.

bryangingechen:
> For once, we cannot distinguish explicitly named a arguments from arrows (at least before leanprover-community/lean#442).

Following up on this, I still think things are better after this change, but there are now a bunch of `ᾰ` in the docs:

https://leanprover-community.github.io/mathlib_docs/meta/expr.html#nat.mk_numeral

robertylewis:
Oops. I had tested this against the default (older) mathlib. Fix incoming once I've tested it for real...

PR #59, opened 2020-08-30
Flashy background highlighting.

gebner:
Patrick doesn't like my color scheme though.

flashy_highlighting

### DIFF
--- a/style.css
+++ b/style.css
@@ -365,16 +365,20 @@ label[for="nav_toggle"] { /* The MENU button */
 }
 
 .def .fn:hover {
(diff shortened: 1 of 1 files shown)


robertylewis:
I love it!

gebner:
How about this?  I think this represents the spirit of Lean very well:

flashy4

robertylewis:
What about a gradient based on the sidebar color (green for theorems, etc)? Or is that too matchy? I guess we have blue links for everything so it could clash.

I'm fine with the current one btw, Patrick and I must have different tastes.

jcommelin:
I like the fact that there is highlighting. I'm too much of a barbarian to have an opinion on the details.

PR #65, opened 2020-09-17
Use CSS for line-breaking in definition names rather than invisible characters

### LINKED ISSUE #58: Do not use zero width space
@rwbarton suggested this on Zulip:
> https://jsfiddle.net/vqwpjL9c/9/ is better

alexpeattie:
Emulate the old line break behaviour, but with CSS rather than zero-width spaces 🎨 . Closes #58

### DIFF
--- a/templates/module.j2
+++ b/templates/module.j2
@@ -1,5 +1,5 @@
 {% extends "base.j2" %}
-{% block title %}{{ filename | filename_import | htmlify_name }}{% endblock %}
+{% block title %}{{ filename | filename_import | e }}{% endblock %}
 
 {% block content %}
 {% for item in items %}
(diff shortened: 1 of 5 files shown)


bryangingechen (approved):
Thank you! I tested this out locally and it looks great.

templates/module.j2
```diff
@@ -1,5 +1,5 @@
 {% extends "base.j2" %}
-{% block title %}{{ filename | filename_import | htmlify_name }}{% endblock %}
+{% block title %}{{ filename | filename_import | e }}{% endblock %}
```
gebner, line 2: Unfortunately we also need linebreaks in module names.  See e.g. this overly long name here: https://leanprover-community.github.io/mathlib_docs/category_theory/limits/shapes/constructions/limits_of_products_and_equalizers.html

I would suggest to have `htmlify_name` return a `<span class="name">....</span>` and replace `.decl_name > a` by `.name` in the CSS.

alexpeattie: Ah thanks @gebner - I hadn't thought about breaks in module names 🙂 . I just pushed a fix

PR #66, opened 2020-09-17
Go back to using regular Pygments package

alexpeattie:
Now Pygments latest releases supports Lean 3, we don't need to use the prerelease version any more 🙂


bryangingechen:
Thanks!

PR #71, opened 2020-09-23
Add meta tags to generate nicer social media previews etc.

alexpeattie:
See discussion here: 

https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/tweets.20mentioning.20the.20docs

Short (200 character) descriptions are generated automatically for regular documentation pages, by taking the initial docstring, stripping out the markdown formatting and truncating it.

For "special" pages (i.e. everything under "General documentation" in the sidebar) I've explicitly added a description. Here's a preview of how the Twitter cards should look:

Screenshot 2020-09-23 at 21 28 02

Since I had to put the meta thumbnail image somewhere, I made a `static/` directory too - everything in `static/` is copied as-is to the `html_root`.


templates/attributes.j2
```diff
@@ -2,6 +2,11 @@
 
 {% block title %}Attributes{% endblock %}
 
+{%- block metadesc -%}
+  {{ "Attributes are a tool for associating information with declarations. 
+  In the simplest case, an attribute is a tag that can be applied to a declaration. simp is a common example of this." | plaintext_summary }}
```
gebner, line 7: Is there a reason to do `plaintext_summary` here?  Wouldn't it be easier to write the result directly?

gebner (commented):
Looks great!  There is no way to make https://leanprover-community.github.io/mathlib_docs/data/complex/exponential.html#complex.exp' show the declaration doc string, right?

templates/attributes.j2
alexpeattie: > Looks great! There is no way to make https://leanprover-community.github.io/mathlib_docs/data/complex/exponential.html#complex.exp' show the declaration doc string, right?

Thanks 😃 ! Unfortunately not - the meta tags have to be per page, not per anchor tag

alexpeattie:
> Looks great! There is no way to make https://leanprover-community.github.io/mathlib_docs/data/complex/exponential.html#complex.exp' show the declaration doc string, right?

Thanks 😃 ! No unfortunately the meta tags have to be per page, not per anchor tag...

templates/attributes.j2
alexpeattie: No real reason - just that it does the truncating to < 200 characters without me having to count hehe. Happy to change it to just putting the summary in directly

gebner: Ok, if we need to respect hard character limits then it's ok.

PR #73, opened 2020-09-24
change highlight colors to match boxes

robertylewis:
(This is a PR to the `flashy_highlighting` branch from #59)

It duplicates some code (I'm not good enough with CSS to know how to avoid it), but the highlight colors are now based on the colors of the surrounding boxes:

highlight1
highlight2

For deeply nested things like the second, the colors do diverge a bit, but I don't think it looks so bad.

### DIFF
--- a/style.css
+++ b/style.css
@@ -365,16 +365,20 @@ label[for="nav_toggle"] { /* The MENU button */
 }
 
 .def .fn:hover {
-    background-color: hsla(187, 61%, calc(100% - 4%*var(--fn)));
+    background-color: hsla(187, 61%, calc(100% - 5%*var(--fn)));
+    border-radius: 5px;
 }
 .theorem .fn:hover {
-    background-color: hsla(115, 62%, calc(100% - 4%*var(--fn)));
+    background-color: hsla(115, 62%, calc(100% - 5%*var(--fn)));
+    border-radius: 5px;
 }
 .axiom .fn:hover, .constant .fn:hover {
-    background-color: hsla(16, 94%, calc(100% - 4%*var(--fn)));
+    background-color: hsla(16, 94%, calc(100% - 5%*var(--fn)));
+    border-radius: 5px;
 }
 .structure .fn:hover, .inductive .fn:hover {
-    background-color: hsla(40, 98%, calc(100% - 4%*var(--fn)));
+    background-color: hsla(40, 98%, calc(100% - 5%*var(--fn)));
+    border-radius: 5px;
 }
 
 .decl_args, .decl_type {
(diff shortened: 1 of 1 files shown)


style.css
```diff
@@ -363,8 +363,18 @@ label[for="nav_toggle"] { /* The MENU button */
 .fn {
     transition: background-color 100ms ease-in-out;
 }
-.fn:hover {
-    background-color: hsla(calc(75 - var(--fn) * 10), 80%, calc(100% - 3%*var(--fn)));
+
+.def .fn:hover {
+    background-color: hsla(calc(187 - var(--fn) * 10), 61%, calc(100% - 3%*var(--fn)));
```
eric-wieser, line 368: Should `--fn` start as `0` instead of `1`, so that the color matches exactly at the top level?

robertylewis: Yes, I think it's a slight improvement.

eric-wieser: Can you update the screenshot?

style.css
```diff
@@ -363,8 +363,18 @@ label[for="nav_toggle"] { /* The MENU button */
...
+}
+.theorem .fn:hover {
+    background-color: hsla(calc(115 - (var(--fn) - 1) * 10), 62%, calc(100% - 3%*var(--fn)));
+}
+.axiom .fn:hover, .constant .fn:hover {
+    background-color: hsla(calc(16 - (var(--fn) - 1) * 10), 94%, calc(100% - 3%*var(--fn)));
+}
+.structure .fn:hover, .inductive .fn:hover {
+    background-color: hsla(calc(40 - (var(--fn) - 1) * 10), 98%, calc(100% - 3%*var(--fn)));
 }
```
eric-wieser, line 378: I wonder if this would look better without the hue variance,
```suggestion
.def .fn:hover {
    background-color: hsla(calc(187, 61%, calc(100% - 3%*var(--fn)));
}
.theorem .fn:hover {
    background-color: hsla(calc(115, 62%, calc(100% - 3%*var(--fn)));
}
.axiom .fn:hover, .constant .fn:hover {
    background-color: hsla(calc(16, 94%, calc(100% - 3%*var(--fn)));
}
.structure .fn:hover, .inductive .fn:hover {
    background-color: hsla(calc(40, 98%, calc(100% - 3%*var(--fn)));
}
```

eric-wieser: image

robertylewis: I was skeptical about this since I thought it would blend together too much, but actually, with the dynamics of the colors darkening as you move your mouse, I think this is the best solution. I even prefer it with 4% steps instead of 5%.

eric-wieser: It would be cool if we could count up in luminosity from the hovered expression, instead of counting down from the ancestor. I can't think of a way to do that with just CSS though.

PR #80, opened 2020-09-28
print localized notation

robertylewis:
As per the discussion here: https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/local.20notation.20in.20doc-gen

This lets us print declarations using (most) localized notation, by opening all locales.

Two caveats:
* Some local notation is malformed, using abbreviated names. This doesn't work without the right namespaces open. Right now the tactic skips any notation it can't open, but this should be fixed in mathlib.
* `lean --run` executes `main` in a different parser setting, so the local notation is no longer in scope. I had to change the setup from main to a run_cmd, which means it tries to run in VSCode, which is super annoying. Not sure what the better fix is here.

image


robertylewis:
Ah, a way to do it with `--run`: un-localize the local notation! I think this behaves differently in the case of overloads, see the difference with `dist`:
image

gebner:
IMHO `has_dist.dist` is the more annoying part.  How about putting the `run_cmd` (as well as the `import all`) into a different file?

robertylewis:
Yeah, that seems to work great.

src/export_json.lean
```diff
@@ -382,6 +382,14 @@ pure $ json.object [
   ("instances", instl)
 ]
 
+open lean.parser
+@[user_command]
+meta def open_all_locales (_ : interactive.parse (tk "open_all_local")): lean.parser unit :=
+do m ← of_tactic localized_attr.get_cache,
+   cmds ← of_tactic $ get_localized m.keys,
+   cmds.mmap' lean.parser.emit_code_here <|> skip
```
gebner, line 390:
```suggestion
   cmds.mmap' lean.parser.emit_code_here
```
This would skip all locales after the first error.

PR #95, opened 2020-11-04
use full mathlib commit in build

robertylewis:
This is why my (accidentally pushed to master) fix doesn't work as expected in production!


bryangingechen (approved):
LGTM

Issue #96, opened 2020-11-16
minisearch webworker is terminates on page reloads


eric-wieser:
I've found the startup time for the search to be very painful, especially since when I click a search result and find its the wrong one, I have to start all over again waiting for it to load.

What's happening here is that the `SharedWorker` is being killed by chrome, because at the point I click the link there are no mathlib_docs tabs open with a search connection to keep it alive.

As a workaround, I can just open an extra tab and start a search in it, which is enough to keep the search alive. Better solutions might be to:
* Intercept internal page navigation and use the javascript history API to replace the page content without actually unloading the page
* Make it so I can open the search results in a new tab.

eric-wieser:
One option for improving this would be to  :

* use `miniSearch.toJSON()` to export the complete index after constructing it for the first time
* put it in `localStorage`
* next tmie a page is loaded, try using `MiniSearch.fromJS(the_json_from_local_storage)`, which would bypass the long startup times

We'd need to make sure there's some mechanism to invalidate that cache when the docs site updates.

PR #97, opened 2020-11-16
Make it possible to open search results in a new tab

eric-wieser:
I ~~haven't tested this (I'm not sure how to)~~ tested this by meddling with the chrome inspector, and I think this would address some of the pain of #96.

By making these proper `a` tags, I can now open results in a new window with either Ctrl+click, middle click, or the context menu.

Hitting enter on a list item works the same way as before.

### DIFF
--- a/nav.js
+++ b/nav.js
@@ -181,7 +181,7 @@ searchInput.addEventListener('input', async (ev) => {
     const d = sr.appendChild(document.createElement('a'));
     d.innerText = decl;
     d.title = decl;
-    d.href = () => `${siteRoot}find/${decl}`;
+    d.href = `${siteRoot}find/${decl}`;
   }
   sr.setAttribute('state', 'done');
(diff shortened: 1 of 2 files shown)


nav.js
```diff
@@ -180,10 +178,10 @@ searchInput.addEventListener('input', async (ev) => {
   const oldSR = document.getElementById('search_results');
   const sr = oldSR.cloneNode(false);
   for (const {decl} of result) {
-    const d = sr.appendChild(document.createElement('div'));
+    const d = sr.appendChild(document.createElement('a'));
     d.innerText = decl;
     d.title = decl;
-    d.onclick = () => goToDecl(decl);
+    d.href = () => `${siteRoot}find/${decl}`;
```
eric-wieser, line 184:
```suggestion
    d.href = `${siteRoot}find/${decl}`;
```

gebner:
Thanks!

PR #99, opened 2020-11-18
Do not write json to stdout

eric-wieser:
Otherwise it can't be used with projects that emit diagnostics

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -391,9 +391,14 @@ do m ← of_tactic localized_attr.get_cache,
         (tok.length = 1 ∧ tok.front.is_alphanum) ∨ tok ∈ ["ε", "φ", "ψ", "W_", "σ", "ζ"]) $
     lean.parser.emit_code_here m <|> skip
 
+def write_file (fn : string) (cnts : string) (mode := io.mode.write) : io unit :=
+do h ← io.mk_file_handle fn io.mode.write,
+   io.fs.write h cnts.to_char_buffer,
+   io.fs.close h
+
 meta def main : io unit := do
 json ← run_tactic mk_export_json,
-put_str json.unparse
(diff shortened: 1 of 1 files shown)


eric-wieser:
This worked fine in my own repository - any idea why it wouldn't be working fine here?

gebner:
> This worked fine in my own repository - any idea why it wouldn't be working fine here?

Unfortunately, we cannot use this for mathlib since `string.to_char_buffer` is excruciatingly slow.  The easiest fix would be to add an io function to lean that can write a string (instead of a char_buffer).

eric-wieser:
So writing to stdout using a different mechanism to writing to a file handle?

src/export_json.lean
```diff
@@ -391,9 +391,14 @@ do m ← of_tactic localized_attr.get_cache,
         (tok.length = 1 ∧ tok.front.is_alphanum) ∨ tok ∈ ["ε", "φ", "ψ", "W_", "σ", "ζ"]) $
     lean.parser.emit_code_here m <|> skip
 
+def write_file (fn : string) (cnts : string) (mode := io.mode.write) : io unit :=
+do h ← io.mk_file_handle fn io.mode.write,
+   io.fs.write h cnts.to_char_buffer,
```
eric-wieser, line 396:
```suggestion
   io.fs.put_str h cnts,
```

gebner:
Unfortunately, `put_str` also converts to `char_buffer` under the hood.

eric-wieser:
Would chunking the conversion help here?

eric-wieser:
I worked around this downstream by just calling `tail -n 1 export.json`, which does the job of separating lean diagnostics from json output. It's ugly, but far easier that digging around in the lean IO internals

PR #100, opened 2020-11-18
Make it possible to call `main` from another python script

eric-wieser:
This makes it possible for me to write a script that does`
```python
import print_docs
del print_docs.extra_doc_files[:]  # these are mathlilb specific
print_docs.main()
```
Obviously this is a hack, but the hack goes in my code not yours :)

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -334,7 +334,7 @@ def mk_site_tree_core(filenames, path=''):
 
   return entries
 
-def setup_jinja_globals(file_map, loc_map):
+def setup_jinja_globals(file_map, loc_map, instances):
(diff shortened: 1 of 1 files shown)


print_docs.py
```diff
@@ -334,7 +334,7 @@ def mk_site_tree_core(filenames, path=''):
 
   return entries
 
-def setup_jinja_globals(file_map, loc_map):
+def setup_jinja_globals(file_map, loc_map, instances):
```
gebner, line 337: Oh I didn't realize this was a global.

gebner:
Thanks! (just waiting for CI to clear)

PR #102, opened 2020-11-19
Overhaul path resolution to allow using non-mathlib projects too

eric-wieser:
Rather than using heuristics to detect which project files belong too, this now queries `lean --path`, and looks for `leanpkg.toml` files in each path directory to assign a project name to each file.

This also means that files from the lean standard library are now shown as `core / init.logic` instead of `core.init.logic`.

Unrelatedly, this prevents this script from trying to create illegal filenames on windows.

---

The sidebar when run against my project now looks like:

image

I'm able to successfully use this in CI with https://github.com/pygae/lean-ga/blob/master/.github/workflows/lean_doc.yml, which produces https://github.com/pygae/lean-ga-docs

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -359,7 +359,7 @@ def mk_site_tree_core(filenames, path=[]):
   for dirname in sorted(set(dirname for dirname, *rest in filenames if rest != [])):
     new_path = path + [dirname]
     entries.append({
-      "kind": "project" if path == '' else "dir",
+      "kind": "project" if not path else "dir",
       "name": dirname,
       "path": '/'.join(new_path[1:]),
       "children": mk_site_tree_core([rest for dn, *rest in filenames if rest != [] and dn == dirname], new_path)
--- a/print_docs.py
+++ b/print_docs.py
@@ -125,7 +125,11 @@ class ImportName(NamedTuple):
         pass
       else:
         return cls(name, rel_path.with_suffix('').parts, fname)
-    raise RuntimeError(f"Cannot locate import name for {fname}")
+    path_details = "".join(f" - {p.raw_path}\n" for p, _ in path_info)
+    raise RuntimeError(
+      f"Cannot determine import name for {fname}; it is not within any of the directories returned by `lean --path`:\n"
+      f"{path_details}"
+      f"Did you generate `export.json` using a different Lean installation to the one this script is running with?")
 
   @property
   def name(self):
--- a/print_docs.py
+++ b/print_docs.py
@@ -3,8 +3,6 @@
 # Requires the `markdown2` and `toml` packages:
 #   `pip install markdown2 toml`
 #
-# This script is not Windows friendly.
-#
 
 import json
 import os
@@ -23,6 +21,9 @@ from urllib.parse import quote
 from functools import reduce
 import textwrap
 from collections import defaultdict
+from pathlib import Path
+from typing import NamedTuple, List
+import sys
 
 root = os.getcwd()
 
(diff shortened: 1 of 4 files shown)


jcommelin:
I like what the end result looks like, but I don't think I should be the one reviewing the python code.

gebner:
1. The links on the sidebar are missing the `.html` extension.
2. I think I'd prefer `.nav h4 { margin-bottom: 1ex; }`

gebner:
3. Also the first time I ran it (just the python script), I got an error that it couldn't locate the import name for `../lean-3.20.0/...`.  It would be nice for testing if this didn't fail on unknown roots (and maybe just printed a warning instead).

eric-wieser:
> 1\.The links on the sidebar are missing the .html extension.

The github-pages webserver seems to add this automatically. Would you like the extension anyway?

> 2\. I think I'd prefer `.nav h4 { margin-bottom: 1ex; }`

I didn't really think about the style at all here to be honest. Feel free to change it as you like. I'll add those lines to make it easier for you to do so on the web

> \3. Also the first time I ran it (just the python script), I got an error that it couldn't locate the import name for `../lean-3.20.0/...`. It would be nice for testing if this didn't fail on unknown roots (and maybe just printed a warning instead).

Right - with this change, the python script and the export.json have to be run on the same machine. An alternate approach would be to save the output of `leanproject.path` in a separate `export-path.json` file, and have the python script read in both. I can't see any easy way to get the path information from within lean, but it would be trivial to have the bash script invoke both.

I think failing on unknown roots is the right thing to do - it means that something is horribly wrong with your lean path, and you should sort it out. I could make the error message more informative if that would help.

gebner:
> The github-pages webserver seems to add this automatically. Would you like the extension anyway?

Wait, it modifies the html?  In any case, the links should be correct for local testing.  (I typically run `python3 -m http.server` in the `html` directory.)

> I think failing on unknown roots is the right thing to do - it means that something is horribly wrong with your lean path, and you should sort it out. I could make the error message more informative if that would help.

My main use case is testing of doc-gen itself.  I usually keep an old `export.json` around because it takes ages to generate.  I think the error message is already good enough.

eric-wieser:
> Wait, it modifies the html?

No, it's part of the apache/nginx/whatever configuration that maps urls to filenames, and that mapping doesn't care about file extensions. Presumably `http.server` is not trying to be as clever.

Keeping it as is has the advantage of shorter URLs. Changing it makes your local workflow easier. I don't care either way, the change was mostly accidental.

eric-wieser:
> I usually keep an old export.json around because it takes ages to generate

Ah, I think I'm thinking of a different problem then. I pulled an export.json from CI and obviously it didn't work on my local machine. Your issue is that your export.json was generated using an older lean version. I suppose it would be reasonable safe to continue with a warning in that case, but the only way I can think of doing that is:

* Add back some heuristics
* Invoke `lean --path > export-path.json` ahead of time, and have you keep that file around too.

gebner:
> No, it's part of the apache/nginx/whatever configuration that maps urls to filenames, and that mapping doesn't care about file extensions. Presumably http.server is not trying to be as clever.

Ok, I understand.  Please just keep the current link urls (with the `.html` at the end).

> Add back some heuristics

Please don't put too much effort into it.  If I get a warning and all files show up in the `unknown` project or something, that would already be enough.

print_docs.py
```diff
@@ -303,28 +353,28 @@ def write_pure_md_file(source, dest, name, loc_map):
...
 
   for filename in sorted(filename for filename, *rest in filenames if rest == []):
-    new_path = filename if path == '' else path+'/'+filename
+    new_path = path + [filename]
     entries.append({
       "kind": "file",
-      "name": filename[:-5],
-      "path": new_path
+      "name": filename,
+      "path": '/'.join(new_path[1:]),
```
eric-wieser, line 377:
```suggestion
      "path": '/'.join(new_path[1:]) + '.html',
```

gebner:
Thanks!

eric-wieser:
> If I get a warning and all files show up in the unknown project or something, that would already be enough.

It wouldn't be too much work to just make it use the full file path as the import path and generate something nasty like `Users.eric.projects.lean.mathlib.src.algebra.algebra.basic` if it can't find the mathlib path, for instance.

Perhaps a better solution to your testing pain is to publish `exports.json` in the docs webpage too, and then you can just download it to get an up to date version?

PR #103, opened 2020-11-30
Add an "imported by" section

eric-wieser:
This makes things a little more painful to run locally, since now you need the full lean environment available to construct the graph.

Live here: https://pygae.github.io/lean-ga-docs/init/default.html

image

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -125,7 +125,7 @@ class ImportName(NamedTuple):
         pass
       else:
         return cls(name, rel_path.with_suffix('').parts, fname)
-    path_details = "".join(f" - {p.raw_path}\n" for p, _ in path_info)
+    path_details = "".join(f" - {p}\n" for p, _ in path_info)
     raise RuntimeError(
       f"Cannot determine import name for {fname}; it is not within any of the directories returned by `lean --path`:\n"
       f"{path_details}"
--- a/print_docs.py
+++ b/print_docs.py
@@ -25,6 +25,8 @@ from pathlib import Path
 from typing import NamedTuple, List
 import sys
 
+import networkx as nx
+
 root = os.getcwd()
 
 from jinja2 import Environment, FileSystemLoader, select_autoescape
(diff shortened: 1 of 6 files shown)


gebner:
1. There is now a `.` project for doc-gen, this seems to be an accident.

2. At first I didn't find the new sections for the imports at all.  I would have never guessed to look at the end of the page (which can be very long).  I think it would be better to have the import information after the first module doc / before the first declaration.  And maybe use `<details>` so that it doesn't take up any space by default.

eric-wieser:
> There is now a . project for doc-gen, this seems to be an accident.

Whoops.

> I think it would be better to have the import information after the first module doc / before the first declaration.

Which rule takes priority? The first module doc can occur after the first declaration

gebner:
> > I think it would be better to have the import information after the first module doc / before the first declaration.
> 
> Which rule takes priority? The first module doc can occur after the first declaration

I'd say whatever comes first.  The imports should come as early as possible, but just not before the first module doc.

eric-wieser:
I'm tempted to say the import information can come before the module docstring, because:

* It makes jumping through multiple layers of imports easy without scrolling
* It matches where they appear in the actual lean files
* If they're collapsed by default, they don't use much space
* I don't fancy fighting with jinja right now to work out how best to do what you suggest!

I've pushed an update which uses `<details>` and puts the imports at the top. When CI finishes here, it should be done at my repo too (if not before)

gebner:
I find it visually very unappealing to put text before an `h1` heading.  I can play with the jinja2 side myself if you don't want to do that.  Would it be better to (also) display the import info in the right nav column?

BTW, what happened to the "Additional documentation" on the left side?

eric-wieser:
> BTW, what happened to the "Additional documentation" on the left side?

That's disabled on my project, because the files it's looking for are in mathlib not my project.

eric-wieser:
> I find it visually very unappealing to put text before an `h1` heading. I can play with the jinja2 side myself if you don't want to do that. Would it be better to (also) display the import info in the right nav column?

Please do whatever you like to the formatting - I just wanted to build the part to extract the information. I thought about using the sidebar too.

gebner:
> > BTW, what happened to the "Additional documentation" on the left side?
> 
> That's disabled on my project, because the files it's looking for are in mathlib not my project. It's not a consequence of this diff, but of my CI setup that monkey-patches some mathlib-specific config away.

Ok!  Feel free to PR a flag to disable mathlib-specific stuff if you'd prefer that.

eric-wieser:
> Ok! Feel free to PR a flag to disable mathlib-specific stuff if you'd prefer that.

I don't think it's worth it to me to do that, what I have keeps doc-gen cleaner, and is only a few lines of hack:

https://github.com/pygae/lean-ga/blob/3027ba22a5698d0d2a9e4f497399a55027704aa9/.github/workflows/lean_doc.yml#L71-L73

gebner:
> > I find it visually very unappealing to put text before an `h1` heading. I can play with the jinja2 side myself if you don't want to do that. Would it be better to (also) display the import info in the right nav column?
> 
> Please do whatever you like to the formatting - I just wanted to build the part to extract the information. I thought about using the sidebar too.

I've now moved it the nav bar.  Putting the imports after the first module doc string doesn't always work either.  Often the first module doc string already contains the first section heading.

eric-wieser:
I just triggered a re-run of my CI, in case that's easier for you to look at.

eric-wieser:
Excellent, thanks!

PR #106, opened 2021-01-17
Improve the tokenizer used by the search engine

sorawee:
The default tokenizer used by minisearch splits queries on
any space and punctuation. However, punctuation is significant in Lean.
E.g., `induction'` and `induction` are different.
By using the default tokenizer, a query like `induction'` will not display
`tactic.interactive.induction'`, but instead mostly show results about
`induction`.

This PR supplies a custom tokenizer which splits on any space (\p{Z})
and dot (.). I added dot as a special case to make the search fuzzy.
For example, searching `tactic.induction` probably should show
results of `tactic.interactive.induction`. Without splitting on dot,
the default fuzzy search would find "tactic.interactive.induction" and
"tactic.induction" too different.

The regex is taken from the page in the comment. Note that the default
tokenizer does exactly the same (see
https://github.com/lucaong/minisearch/blob/fa2e104fb7c26fd7f47db7be6a364ec1693ee606/src/MiniSearch.ts#L1169)

### DIFF
--- a/searchWorker.js
+++ b/searchWorker.js
@@ -7,13 +7,13 @@ const declNames = req.responseText.split('\n');
 
 // Adapted from the default tokenizer and
 // https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BZ%7D&abb=on&c=on&esc=on&g=&i=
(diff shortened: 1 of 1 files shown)


robertylewis:
@sorawee thanks!

@gebner this looks reasonable to me, but the search engine was your doing, what do you think?

gebner:
LGTM.

Issue #62, opened 2020-09-12
Latex Rendering issue


shingtaklam1324:
https://leanprover-community.github.io/mathlib_docs/data/mv_polynomial/basic.html

![image](https://user-images.githubusercontent.com/15098580/92995346-67b1d180-f535-11ea-8083-2b0c4e1abc47.png)

The source is

```
might denote $R[X_i : i \in \sigma]$. It is the type of multivariate
```

which looks like valid Latex to me.

digama0:
The source on the website says
```
might denote $R[X_i : i \in sigma]$. It is the type of multivariate
```
so this is not mathjax's fault.

bryangingechen:
We should investigate and fix the underlying issue, but for this particular example I'm surprised we didn't just use the unicode `σ`.

alexpeattie:
Looks like a weird `markdown2` bug. Minimal reproduction:

```python
>>> import markdown2

>>> markdown2.markdown('Example: $x \in \sigma$ and separately `y`')
'<p>Example: $x \\in \\sigma$ and separately <code>y</code></p>\n'

>>> markdown2.markdown('Example: $x \in \sigma$ and separately `s`')
'<p>Example: $x \\in sigma$ and separately <code>s</code></p>\n'

>>> markdown2.markdown('Example: $x \in \sigma$ and separately `sigm`')
'<p>Example: $x \\in sigma$ and separately <code>sigm</code></p>\n'

>>> markdown2.markdown('Example: $x \in \sigma$ and separately `simg`')
'<p>Example: $x \\in \\sigma$ and separately <code>simg</code></p>\n'

>>> markdown2.markdown('Example: separately `s` and $x \in \sigma$')
'<p>Example: separately <code>s</code> and $x \\in sigma$</p>\n'
```

Escaping in a backslash escaping seems to break if there's a `<code>` tag before or afterwards containing a prefix of the escaped word 🙈 . I can't see any reason why that would be. In this example it's because of the ``s`` [here](https://github.com/leanprover-community/mathlib/blob/d5be9f388a19f891daed609add98153b3024c100/src/data/mv_polynomial/basic.lean#L51), remove that and the escaping on `\sigma` works as expected.

alexpeattie:
I've tracked down a few other places this is happening:

[...]

PR #110, opened 2021-02-14
switch from markdown2 to mistletoe

bryangingechen:
markdown2 has a lot of trouble with latex+mathjax, see e.g. #10, #62 and various issues on Zulip. mistletoe is what leanprover-community.github.io uses and it seems to do better with these examples.

As part of this effort, I copied some of the implementation of Markdown + LaTeX from StackExchange's code math.stackexchange and mathoverflow. I believe #10 and #62 are now fixed, though some workarounds that were committed are now broken, e.g. the module doc string right before `formal_multilinear_series.change_origin` in `analysis.analytic.basic`. Compare [mathlib_docs](https://leanprover-community.github.io/mathlib_docs/analysis/analytic/basic.html#formal_multilinear_series.change_origin) to [mathlib_docs_demo](https://leanprover-community.github.io/mathlib_docs_demo/analysis/analytic/basic.html#formal_multilinear_series.change_origin).

For a bunch of LaTeX examples (including ones copied from the issues above), see [the new test page, latex.html](https://leanprover-community.github.io/mathlib_docs_demo/latex.html).

Since I was making a bunch of changes to the markdown rendering anyways, I went ahead and factored it all out into a separate file.

Some smaller changes:

- Markdown headings now have a `#` link after them, which allows linking to the middle of a page (as in https://github.com/leanprover-community/leanprover-community.github.io/pull/11).

- The syntax highlighting now defaults to `lean`.

- Some dead code was removed.

---

I plan to follow this up with a PR to our naming conventions with some advice on LaTeX + Markdown, now that things should be more consistent.

### DIFF
--- a/mistletoe_renderer.py
+++ b/mistletoe_renderer.py
@@ -1,0 +1,112 @@
+"""
+This module contains a class CustomHTMLRenderer, which uses
+mistletoe to generate HTML from markdown.
+
+Extra features include:
+- Library note links
+- Managing LaTeX so that MathJax will be able to process it in the browser
+- Syntax highlighting with Pygments
+"""
+import re
+
+from mistletoe import Document, HTMLRenderer, span_token
+from pygments import highlight
+from pygments.lexers import get_lexer_by_name as get_lexer
+from pygments.formatters.html import HtmlFormatter
+
+from mathjax_editing import remove_math, replace_math
+
+
+class NoteLink(span_token.SpanToken):
+    """
+    Detect library note links
+    """
+    parse_inner = False
+    pattern = re.compile(r'Note \[(.*)\]', re.I)
+
+    def __init__(self, match):
+        self.body = match.group(0)
+        self.note = match.group(1)
+
+
+class CustomHTMLRenderer(HTMLRenderer):
+    """
+    Call the constructor with `site_root`.
+
+    The main rendering function is `render_md`.
+    """
+
+    def __init__(self, site_root):
+        self.site_root = site_root
+        super().__init__(NoteLink)
+
+    def render_md(self, ds):
+        """
+        A wrapper for this class's .render() function.
+
+        Input is a string containing markdown with LaTeX,
+        Output is a string containing HTML.
+
+        Uses `mathjax_editing` to strip out sections of the text
+        which potentially contain LaTeX and then splice them back in.
+        """
+        ds_no_math, math = remove_math(ds, '$')
+        # We have to run `mathjax_editing.replace_math` on the text in code
+        # blocks before passing it to Pygments (see `render_block_code`),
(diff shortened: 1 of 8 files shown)


bryangingechen:
I'm going to try to modify the markdown processor so that it's as close to the behavior on https://math.stackexchange.com and https://mathoverflow.net as possible, since (in my opinion) those sites are the most significant places on the web right now where users frequently input LaTeX embedded in markdown.

This is likely to break some of the LaTeX we currently have, but I think the consistency will be worth it.

robertylewis:
@bryangingechen thank you for taking this on! This change is definitely overdue.

I'm happy to merge this as soon as there's no obvious breakage worse than what it fixes. I see you marked this as a draft, is there more you're planning to change?

gebner:
Ok, let's try this out!

#deploy :octopus:

robertylewis:
https://github.com/leanprover-community/doc-gen/actions/runs/568726686 :crossed_fingers:

gebner:
Ok, the trigger seems to work!  I've canceled the workflow, because I realized that we still need to merge master here in order to get the new `deploy_docs.sh`

gebner:
#deploy

robertylewis:
Okay, this failed because `leanprover-community-bot` didn't accept the invitation to the demo repo. Trying again:

#deploy

bryangingechen:
Oh, this new test workflow is cool!

> I'm happy to merge this as soon as there's no obvious breakage worse than what it fixes. I see you marked this as a draft, is there more you're planning to change?

Yeah, I still haven't finished the changes described in my second comment above. The good news is I think I've already managed to fix #10. I'll try to clean things up and push a commit with my progress tonight.

gebner:
#deploy and fix the $4

bryangingechen:
#deploy 🙏

mistletoe_renderer.py
```diff
@@ -0,0 +1,112 @@
...
+        which potentially contain LaTeX and then splice them back in.
+        """
+        ds_no_math, math = remove_math(ds, '$')
+        # We have to run `mathjax_editing.replace_math` on the text in code
+        # blocks before passing it to Pygments (see `render_block_code`),
+        # otherwise `replace_math` will be confused by the added syntax
+        # highlighting `<span>`s and won't be able to splice in those blocks.
+        self.math = math
+        html = self.render(Document(ds_no_math))
+        return replace_math(html, self.math)
```
eric-wieser, line 60: I don't quite follow why this is necessary - does `mistletoe.latex_token.Math` not work here?

bryangingechen: I posted a reply [here](https://github.com/leanprover-community/doc-gen/pull/115#issuecomment-779955840). But let's see how #116 turns out!

PR #111, opened 2021-02-15
add action for deploying demo builds on request

robertylewis:
I had this idea a while back but was inspired by #110 to actually implement it.

This adds a listener for the token `#deploy` in PR comments. If you make a comment with this token, it will build the docs using the PR source and upload them to a demo repo (fixed to `leanprover-community/mathlib_docs_demo` ~~which I haven't created yet~~).

There's a bunch of duplicated code between this and the action we use now. I'm not really sure how to make this more modular. Also this is only half-tested. Any thoughts?

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,0 +1,82 @@
+name: build test site on trigger comment
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: khan/pull-request-comment-trigger@master
+        id: check
+        with:
+          trigger: '#deploy'
+          reaction: rocket
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+
+      # this step is needed to get the PR ref from an issue comment
+      # https://github.com/actions/checkout/issues/331
+      - uses: actions/github-script@v3
+        if: steps.check.outputs.triggered == 'true'
+        id: get-pr
+        with:
+          script: |
+            const request = {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            }
+            core.info(`Getting PR #${request.pull_number} from ${request.owner}/${request.repo}`)
+            try {
+              const result = await github.pulls.get(request)
+              return result.data
+            } catch (err) {
+              core.setFailed(`Request failed with error ${err}`)
+            }
+      - uses: actions/checkout@v2
+        if: steps.check.outputs.triggered == 'true'
+        with:
+          repository: ${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}
+          ref: ${{ fromJSON(steps.get-pr.outputs.result).head.sha }} # or .head.ref for branch name
+
+      - name: Checkout mathlib
+        if: steps.check.outputs.triggered == 'true'
+        run: git clone https://github.com/leanprover-community/mathlib
+
+      - name: install elan
+        if: steps.check.outputs.triggered == 'true'
+        run: |
+          set -o pipefail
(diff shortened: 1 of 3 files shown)


gebner:
This is great!  Do you think there are any security issues when random people can write a deploy comment?

#deploy now!

robertylewis:
People will be able to host arbitrary (static) html at https://leanprover-community.github.io/mathlib_docs_demo but I don't think it's a big deal. This repo is low traffic, we'll notice if anyone's abusing it.

Obviously there will be conflicts if there are two active PRs here at once, but again, I think we can manage.

I'd prefer that the /mathlib_docs_demo directory not be indexed by search engines. We can do this with a `robots.txt` in the main website, right?

gebner:
I'm not worried about write-access to the demo repo.  But the `DEPLOY_GITHUB_TOKEN` also works on the main documentation site, right?  (And maybe more?)

The `robots.txt` should work to prevent indexing.

robertylewis:
I created the repo so this should in principle work. The odds of me getting the deploy logic right on my first try are kind of low, but it's hard to test...

gebner:
Once the potential security issues are resolved, I think we should just merge it and try it out.

robertylewis:
That token is just a leanprover-community-bot PAT, it's already stored in this repo and used in the standard deploy action.

Since the new action listens to `issue_comment` events, GitHub will always run the version on `master`. So PRs modifying this action won't execute it. And the token is only accessible to PRs that come from branches of this repo.

gebner:
> So PRs modifying this action won't execute it.

But you can still modify `deploy_docs.sh`, which gets the token as an environment variable.  Can we at least use a less-priviledged token here?

robertylewis:
Ah, I see what you mean. Hmm.

We can create another bot that only has write access to the demo repo. Or we can change the action script to always use `deploy_docs.sh` from `master`. The latter is safer but maybe uglier, what do you think?

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: run leanproject
+        if: steps.check.outputs.triggered == 'true'
+        run: |
+          cd mathlib
+          leanproject up
+
+      - name: generate docs
+        if: steps.check.outputs.triggered == 'true'
+        run: ./deploy_docs.sh "mathlib" ".." "mathlib" "leanprover-community" "mathlib_docs_demo" "true"
```
robertylewis, line 77:
```suggestion
        run: |
          wget https://raw.githubusercontent.com/leanprover-community/doc-gen/master/deploy_docs.sh
          ./deploy_docs.sh "mathlib" ".." "mathlib" "leanprover-community" "mathlib_docs_demo" "true"
```

@gebner this would solve the security problem you're worried about, right? I think it's reasonable to ask that we not use `#deploy` to test the deployment script itself.

gebner:
`deploy_docs.sh` calls `gen_docs`, which calls pretty much everything we want to test in a PR.  I don't think we can stop the leaks there.

It's probably easiest to just copy the authentication code from mathlib, and only allow admins to use the deploy hashtag.

robertylewis:
But `deploy_docs.sh` doesn't pass the private token to any other script. It's only used in one `git clone`. (Which is called before any local script, so even an exploit that replaces the `git` executable should be ruled out.)

If I'm still overlooking something, though, you're probably right that only allowing admins to deploy is easiest.

gebner:
`DEPLOY_GITHUB_TOKEN` is an environment variable.  Environment variables are passed to child processes.

Even if the environment variable was not inherited, a malicious `src/export_json.lean` script could just read the token from the git repository.

robertylewis:
Okay, you're right that the ability to read the token from the repo makes this impossible to secure.

I've tried to copy the logic from mathlib that checks for admin permissions. (Now I'll make more people admins!)

gebner:
LGTM!

robertylewis:
Hmm, it should probably also reply to the PR with a link once the build is uploaded successfully.

PR #115, opened 2021-02-16
Try to use the builtin math token

eric-wieser:
I didn't see this attempted in #110


robertylewis:
@eric-wieser you should have permission to test with `#deploy` when you're ready.

eric-wieser:
Is `#deploy` linked at all to the CI run? Should I wait for CI to finish before running it, or does it run completely separately anyway?

eric-wieser:
Well, I guess I'll just try

#deploy

eric-wieser:
#deploy

bryangingechen:
> Is `#deploy` linked at all to the CI run? Should I wait for CI to finish before running it, or does it run completely separately anyway?

It runs separately.

eric-wieser:
It looks like it doesn't work because of these lines:

https://github.com/leanprover-community/doc-gen/blob/7a66a7c5454070a0c9ab27a21019b637d86a6550/.github/workflows/build-pr.yml#L31-L33

robertylewis:
Ah, right. I think that's not necessary. Have to run now but I'll look into it later.

bryangingechen:
The `pattern` for `latex_token.Math` is a simple regex found here:

https://github.com/miyuchina/mistletoe/blob/master/mistletoe/latex_token.py#L9

Indeed, I didn't try to use this directly. But I did try a number of more complicated regex patterns that attempted to deal with escaped `\$`, and they didn't work well on the test file either.

eric-wieser:
I've reopened this PR from a non-fork so that I can try CI.

PR #117, opened 2021-02-18
references page and links

bryangingechen:
Adds a `references.html` page, generated from mathlib's `docs/references.bib` and creates links from all strings of the form "[bib_entry_key]" in docstrings to the entry on that page. The keys are replaced with auto-generated labels following [the BibTeX "alpha" style](https://www.bibtex.com/s/bibliography-style-base-alpha/).

Also includes backrefs for both bib entries and library notes (links back from the bib / note entries to the pages that cite them). Some of the entries without backrefs correspond to links in `archive/` (e.g. `Hofstadter-1979`) and some are truly missing (e.g. `joyal1977`, there's a `joyal1997` in `set_theory.pgame` which is probably a typo).

Finally, I added functionality that turns raw URLs in the markdown into links.

I did basically the bare minimum, copying as much as I could from leanprover-community.github.io, so any styling / formatting suggestions are welcome!

Screen Shot 2021-02-20 at 12 11 41 AM

[Zulip thread here](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/https.3A.2F.2Fleanprover-community.2Egithub.2Eio.2Fmathlib_docs.2F/near/226893761).

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -356,30 +356,30 @@ def linkify_markdown(string, loc_map):
   def linkify_note(body, note):
     if note in global_notes:
       return f'<a id="{note_backref(note)}" href="{site_root}notes.html#{note}">{body}</a>'
-    return f'{body}'
-  def linkify_named_ref(name, key):
+    return body
+  def linkify_named_ref(body, name, key):
     if key in bib.entries:
       return f'<a id="{bib_backref(key)}" href="{site_root}references.html#{key}">{name}</a>'
-    return f'[{name}][{key}]'
-  def linkify_standalone_ref(key):
+    return body
+  def linkify_standalone_ref(body, key):
     if key in bib.entries:
-      return f'<a id="{bib_backref(key)}" href="{site_root}references.html#{key}">[{key}]</a>'
-    return f'[{key}]'
+      return f'<a id="{bib_backref(key)}" href="{site_root}references.html#{key}">{body}</a>'
+    return body
 
   # notes
   string = re.sub(note_pattern,
-    lambda p: f'{linkify_note(p.group(0), p.group(1))}', string)
+    lambda p: linkify_note(p.group(0), p.group(1)), string)
   # inline declaration names
   string = re.sub(r'<code>([^<]+)</code>',
-    lambda p: '<code>{}</code>'.format(linkify_type(p.group(1))), string)
+    lambda p: f'<code>{linkify_type(p.group(1))}</code>', string)
   # declaration names in highlighted Lean code snippets
   string = re.sub(r'<span class="n">([^<]+)</span>',
-    lambda p: '<span class="n">{}</span>'.format(linkify_type(p.group(1))), string)
+    lambda p: f'<span class="n">{linkify_type(p.group(1))}</span>', string)
   # references (don't match if there are illegal characters for a BibTeX key, cf. https://tex.stackexchange.com/a/408548)
   string = re.sub(r'\[([^\]]+)\]\s*\[([^{ },~#%\\]+)\]',
-    lambda p: f'{linkify_named_ref(p.group(1), p.group(2))}', string)
+    lambda p: linkify_named_ref(p.group(0), p.group(1), p.group(2)), string)
   string = re.sub(r'\[([^{ },~#%\\]+)\]',
-    lambda p: f'{linkify_standalone_ref(p.group(1))}', string)
+    lambda p: linkify_standalone_ref(p.group(0), p.group(1)), string)
   return string
 
 def plaintext_summary(markdown, max_chars = 200):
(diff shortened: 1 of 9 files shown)


gebner:
#deploy :pray:

print_docs.py
```diff
@@ -299,11 +333,26 @@ def linkify_type(string):
     splitstr = re.split(r'([\s\[\]\(\)\{\}])', string)
     tks = map(lambda s: linkify(s, loc_map), splitstr)
     return "".join(tks)
-
+  def linkify_named_ref(name, string):
+    if string in bib.entries:
```
bryangingechen, line 337: It'd be nice to generate backrefs, i.e. links on `references.html` back to the pages where the citations occur, but I need a way to get the file name here. 

I don't have much experience with jinja templates, so I was thinking of changing `convert_markdown` to a higher order function like this:

```python
def convert_markdown(filename):
  def _convert(string):
    # current body of convert_markdown
    pass
  return _convert
```

And then I'd change all the templates using `convert_markdown` to pass their filenames as an argument. Is there a cleaner way of doing this?

gebner: I don't know about clean, but I'd rather go for global variables.  You'll also need a way to generate ids for every link.

bryangingechen: What did you have in mind with global variables? Do you mean having something like `current_file` in the global scope and then assigning to it each time we start rendering another file?

bryangingechen: OK, got it! For the backrefs, I was thinking of just using a new field for each item in `bib.entries` (which is already global).

bryangingechen:
#deploy

bryangingechen:
Two problems at https://leanprover-community.github.io/mathlib_docs_demo/topology/sheaves/sheaf_condition/opens_le_cover.html#backref1

- There's an extra linebreak before the link to Spectral Algebraic Geometry.
- On Firefox, the link is still covered by the header.

Unfortunately, these both seem to be related to the CSS hack that I don't fully understand yet.

bryangingechen:
#deploy

bryangingechen:
#deploy

bryangingechen:
3 things I want to add before this is merged:
- [ ] fix empty "Referenced by" on notes.html
- [ ] add `title` attribute to backrefs
- [ ] format the BibTeX keys using https://bitbucket.org/pybtex-devs/pybtex/src/master/pybtex/style/labels/alpha.py

There's an active discussion about [Stacks tags on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/stacks.20tags), so that may be the next PR.

bryangingechen:
#deploy

bryangingechen:
#deploy

print_docs.py
```diff
@@ -79,6 +81,63 @@
...
+    # turn LaTeX special characters to Unicode,
+    # since format_label does not correctly abbreviate names containing LaTeX
+    author.last_names = list(map(clean_tex, author.last_names))
+  label = label_style.format_label(data)
+  counted.update([label])
+  data.fields['alpha_label'] = label
+# count will track number of times label has been finalized
+count = Counter()
+
+for key, data in bib.entries.items():
```
bryangingechen, line 107: The ordering of the bib entries here is whatever it is in mathlib's `docs/references.bib`. This could potentially lead to strange ordering of the a, b, c that are appended to labels.

print_docs.py
```diff
@@ -79,6 +81,63 @@
 # root directory of mathlib.
 local_lean_root = os.path.join(root, cl_args.r if cl_args.r else '_target/deps/mathlib') + '/'
 
+bib = pybtex.database.parse_file(f'{local_lean_root}docs/references.bib')
+
+label_style = LabelStyle()
```
bryangingechen, line 86: Using `LabelStyle` here is a bit of a hack since I couldn't figure out how to use pybtex's bibliography formatting classes here. (If someone can figure it out, feel free to fix this!)

bryangingechen:
#deploy

print_docs.py
```diff
@@ -79,6 +81,66 @@
...
+    author.last_names = list(map(clean_tex, author.last_names))
+  label = label_style.format_label(data)
+  counted.update([label])
+  data.alpha_label = label
+# count tracks the number of times a duplicate label has been finalized
+count = Counter()
+
+for key, data in bib.entries.items():
+  label = data.alpha_label
+  # Finalize duplicate labels by appending 'a', 'b', 'c', etc.
```
bryangingechen, line 109:
```suggestion
  # Finalize duplicate labels by appending 'a', 'b', 'c', etc.
  # Currently the ordering is determined by `docs/references.bib`
```

robertylewis:
@bryangingechen this looks fantastic. I'm happy to merge if you are. @gebner ?

gebner:
Yeah, looks great, Bryan!

bryangingechen:
Thanks! I'll merge this now then!

PR #124, opened 2021-03-05
parallelize json export

robertylewis:
This is much faster (obviously) and uses less memory (surprisingly).

Thanks to @gebner for https://github.com/leanprover-community/mathlib/pull/6293

### DIFF
--- a/leanpkg.toml
+++ b/leanpkg.toml
@@ -1,8 +1,8 @@
 [package]
 name = "."
 version = "0.1"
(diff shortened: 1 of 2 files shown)


bryangingechen:
Wow, this takes the build time from ~30 minutes to 8!

Note that @gebner has a PR in progress removing part of the PR linked above after changes in core Lean, but it doesn't seem to build yet: https://github.com/leanprover-community/mathlib/pull/6441

PR #125, opened 2021-03-09
Try to speed up the download of decl.txt

eric-wieser:
If `bmp` doesn't work, `js` might.

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -666,7 +666,9 @@ def copy_static_files(path):
     shutil.copy(filename, path)
 
 def write_decl_txt(loc_map):
-  with open_outfile('decl.txt') as out:
+  # it's not a bitmap, but apparently this tricks Github Pages / nginx into gzipping it
+  # https://github.com/isaacs/github/issues/611#issuecomment-526848190
+  with open_outfile('decl.bmp') as out:
     out.write('\n'.join(loc_map.keys()))
 
 def mk_export_map_entry(decl_name, filename, kind, is_meta, line, args, tp):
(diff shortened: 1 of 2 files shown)


eric-wieser:
#deploy

print_docs.py
```diff
@@ -666,7 +666,9 @@ def copy_static_files(path):
     shutil.copy(filename, path)
 
 def write_decl_txt(loc_map):
-  with open_outfile('decl.txt') as out:
+  # it's not a bitmap, but apparently this tricks Github Pages / nginx into gzipping it
```
gebner, line 669: I'm not sure whether I should be disgusted or impressed.  Either way, if it works let's merge it.

eric-wieser: Would `.js` leave you less disgusted?

searchWorker.js
```diff
@@ -1,5 +1,5 @@
 const req = new XMLHttpRequest();
-req.open('GET', 'decl.txt', false /* blocking */);
+req.open('GET', 'decl.bmp', false /* blocking */);
```
eric-wieser, line 2:
```suggestion
req.open('GET', 'decl.txt.js', false /* blocking */);
```

print_docs.py
```diff
@@ -666,7 +666,9 @@ def copy_static_files(path):
...
+  # https://github.com/isaacs/github/issues/611#issuecomment-526848190
+  with open_outfile('decl.bmp') as out:
```
eric-wieser, line 671:
```suggestion
  with open_outfile('decl.txt.js') as out:
```

print_docs.py
eric-wieser:
```suggestion
  # it's not a javascript file, but apparently this tricks Github Pages / nginx into gzipping it
```

PR #126, opened 2021-03-14
don't embed commit SHA in as many files

bryangingechen:
This PR is intended to solve problems we've had with the `mathlib_docs` repo. Previously, every update of the docs led to tens of thousands of new blobs being created, because the commit SHA was embedded in every doc page and in every src redirect page.

I made the following changes:
- [x] source links in the HTML files of the doc pages now link to `master`, and `nav.js` rewrites the URLs to point to the precise commit when the page loads. To test: [visit any doc page](https://leanprover-community.github.io/mathlib_docs_demo/meta/expr.html) and click a "source" link; it should take you to the right place on GitHub. It may be worth testing both pages in mathlib and in core Lean since GitHub links for the latter aren't rewritten.
- [x] the src redirect pages now do not contain the commit SHA; a separate `add_commit.js` file performs the redirect. To test: try [this src link for `dvd_sub_pow_of_dvd_sub`](https://leanprover-community.github.io/mathlib_docs_demo/find/dvd_sub_pow_of_dvd_sub/src) (in mathlib, requires a rewrite) and [this src link for `nat`](https://leanprover-community.github.io/mathlib_docs_demo/find/nat/src) (in core Lean, no rewrite required).

### DIFF
--- a/print_docs.py
+++ b/print_docs.py
@@ -157,10 +157,35 @@ mathlib_github_root = ml_data['git'].strip('/')
 if cl_args.w:
   site_root = cl_args.w
 
-mathlib_github_src_root = "{0}/blob/{1}/src/".format(mathlib_github_root, mathlib_commit)
+# contains an array of [prefix, replacement] strings to be rewritten by nav.js
+url_rewrites = []
 
+mathlib_github_src_root_with_commit = f"{mathlib_github_root}/blob/{mathlib_commit}/src/"
+mathlib_github_src_root = f"{mathlib_github_root}/blob/master/src/"
+url_rewrites.append([mathlib_github_src_root, mathlib_github_src_root_with_commit])
+
+# The Lean version changes infrequently enough that we don't need to rewrite it
 lean_commit = subprocess.check_output(['lean', '--run', 'src/lean_commit.lean']).decode()
-lean_root = 'https://github.com/leanprover-community/lean/blob/{}/library/'.format(lean_commit)
+lean_root = f'https://github.com/leanprover-community/lean/blob/{lean_commit}/library/'
+
+def modify_nav_js(url_rewrites: List):
+  """nav.js will fetch commit.json and use it to rewrite the href attributes of all links in gh_link"""
+  with open_outfile('nav.js', 'a') as out:
+    out.write(f"commit = {json.dumps(url_rewrites)};")
+    out.write("""
+// Rewrite GitHub links
+// --------------------
+
+for (const elem of document.getElementsByClassName('gh_link')) {
+  const a = elem.firstElementChild;
+  for (const [prefix, replacement] of commit) {
+    if (a.href.startsWith(prefix)) {
+      a.href = a.href.replace(prefix, replacement);
+      break;
+    }
+  }
+}
+""")
 
 def get_name_from_leanpkg_path(p: Path) -> str:
   """ get the package name corresponding to a source path """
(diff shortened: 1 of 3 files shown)


bryangingechen:
For the src redirect pages, I'm thinking of trying out @digama0's idea:

> the `find/polynomial/src` page can have the form:
```html
<script href="add_commit.js"></script>
<script>redirectTo("src/data/polynomial/basic.lean#L24")</script>
<noscript>
<meta http-equiv="refresh" content="0;url=https://github.com/leanprover-community/mathlib/blob/master/src/data/polynomial/basic.lean#L24">
</noscript>
```

> `add_commit.js`:
```js
function redirectTo(tgt) {
  window.href = "https://github.com/leanprover-community/mathlib/blob/<commit>/" + tgt;
}
```

bryangingechen:
#deploy

bryangingechen:
#deploy

print_docs.py
```diff
@@ -157,10 +157,38 @@ def parse_bib_file(fname):
 if cl_args.w:
   site_root = cl_args.w
 
-mathlib_github_src_root = "{0}/blob/{1}/src/".format(mathlib_github_root, mathlib_commit)
+# contains an array of [prefix, replacement] strings to be rewritten by nav.js
+url_rewrites = []
```
bryangingechen, line 161: I'm storing the rewrites in an array like this since it might make it easier for `doc-gen` to be used by projects outside mathlib.

bryangingechen:
#deploy

robertylewis:
Nice! As far as I can tell this works as expected.

With this change, the update diffs should be pretty small. Should we change the push logic (in another commit) to do non-destructive updates?

print_docs.py
```diff
@@ -641,21 +692,25 @@ def write_redirects(loc_map, file_map):
...
-      except FileNotFoundError:
-        pass
       os.symlink(os.path.relpath(a, os.path.dirname(b)), b)
     else:
       shutil.copyfile(a, b)
 
   cp('style.css', path+'style.css')
   cp('pygments.css', path+'pygments.css')
-  cp('nav.js', path+'nav.js')
+  shutil.copyfile('nav.js', path+'nav.js')
```
gebner, line 710: This really breaks my workflow.  When changing `nav.js`, I don't want to wait for `doc-gen` every time.

Could we instead include the `add_commit.js` file before `nav.js` and use the `commit` variable from there?

print_docs.py
```diff
@@ -626,13 +654,36 @@ def write_site_map(partition):
...
     out.write(f'<meta http-equiv="refresh" content="0;url={url}#{quote(decl_name)}">')
 
 def write_src_redirect(decl_name, decl_loc, file_map):
   url = library_link_from_decl_name(decl_name, decl_loc, file_map)
-  with open_outfile('find/' + decl_name + '/src/index.html') as out:
-    out.write(f'<meta http-equiv="refresh" content="0;url={url}">')
+  with open_outfile(f'find/{decl_name}/src/index.html') as out:
+    out.write(f"""<script src="{site_root}add_commit.js"></script>
+<script>redirectTo("{url}");</script>
+<noscript><meta http-equiv="refresh" content="0;url={url}"></noscript>
```
gebner, line 665:
```suggestion
<meta http-equiv="refresh" content="0;url={url}">
```

Does it work without the `<noscript>` tag as well?  `<noscript>` isn't particularly reliable in the age of adblockers.

print_docs.py
bryangingechen: Done!

print_docs.py
bryangingechen: This works, but I'm not sure why. Won't there be some kind of race condition between the javascript and the `<meta>` tag?

bryangingechen:
#deploy

bryangingechen:
> With this change, the update diffs should be pretty small. Should we change the push logic (in another commit) to do non-destructive updates?

We're force pushing because cloning the history was taking too long, right? I think we can try out going back to normal pushes.

print_docs.py
gebner: As a compromise, could we leave the `<noscript>` tag but add a clickable link (`<a>`)?

gebner:
> We're force pushing because cloning the history was taking too long, right? I think we can try out going back to normal pushes.

The other reason was space: the generated documentation is several gigabytes large, and most of the files are changing every time (due to the git revisions).

bryangingechen:
#deploy

PR #127, opened 2021-03-24
swap order of "mathlib docs" and {% block title %}

Madjosz:
Swap order of "mathlib docs" and {% block title %} in HTML <title> to improve readability in browser tabs.

When you have multiple open tabs of mathlib documentation you cannot distinguish them without hovering the tab text to show a toolbox in many browsers:
Tabs looking all the same

I propose to swap the order like it is done in most big websites (e.g. Google uses "title - Google", Facebook uses "title | Facebook").

### DIFF
--- a/templates/base.j2
+++ b/templates/base.j2
@@ -4,18 +4,18 @@
     <link rel="stylesheet" href="{{ site_root }}style.css">
     <link rel="stylesheet" href="{{ site_root }}pygments.css">
     <link rel="shortcut icon" href="{{ site_root }}favicon.ico">
-    <title>mathlib docs: {% block title %}{% endblock %}</title>
+    <title>{% block title %}{% endblock %} - mathlib docs</title>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


bryangingechen (approved):
#deploy

LGTM; let's test it out.

eric-wieser:
Should we change these lines too?

https://github.com/leanprover-community/doc-gen/blob/d146c79484d09703973d4c8b5af25703507d2cea/templates/base.j2#L12
https://github.com/leanprover-community/doc-gen/blob/d146c79484d09703973d4c8b5af25703507d2cea/templates/base.j2#L18

Madjosz:
I would say to keep the other two places as they are used for content previews in e.g. social media. Compare the current state
current preview
to a possible changed state
possible future preview

eric-wieser:
I can't think of a particularly good reason to use a different html title and open-graph / twitter title; it's misleading to have a content preview use a different page title to the page itself, and might even be penalized by SEO.

Madjosz:
Ok, changed these two lines also.  I am not sure what that `{% endblock %}` does in the HTML title.

bryangingechen:
> I am not sure what that `{% endblock %}` does in the HTML title.

This is [Jinja template syntax](https://jinja.palletsprojects.com/en/2.11.x/templates/#template-inheritance). Basically `{% block blah %}{% endblock %}` will insert the `blah` template at that point.

PR #129, opened 2021-04-10
README: clarify line about .olean files

waldyrious:
Just a small tweak to make the line less cryptic for beginners.

Although, to be honest I still don't fully understand what this sentence is saying — _how_ does one make sure the `.olean` files are generated in the `_target` folder, and what exactly will be extremely slow otherwise?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,7 +15,8 @@ rm -rf _target
 leanproject up
 ```
 
-Make sure that olean files are generated for mathlib in `_target`, otherwise this will be extremely slow.
+Make sure that [`.olean` files](https://github.com/leanprover/tutorial/blob/master/05_Interacting_with_Lean.org#projects)
+are generated for mathlib in `_target`, otherwise this will be extremely slow.
 
 ## Usage
 


bryangingechen:
I usually run `leanproject up` to get the latest mathlib master, or simply `leanproject get-mathlib-cache` if I don't care what version of mathlib I'm generating the docs for.

readme.md
```diff
@@ -15,7 +15,8 @@ rm -rf _target
 leanproject up
 ```
 
-Make sure that olean files are generated for mathlib in `_target`, otherwise this will be extremely slow.
+Make sure that [`.olean` files](https://github.com/leanprover/tutorial/blob/master/05_Interacting_with_Lean.org#projects)
```
bryangingechen, line 18: We most likely don't want to link to that tutorial which is very out of date. It's better to link to the community webpage on `leanproject`.

bryangingechen:
> what exactly will be extremely slow otherwise?

If the `.olean` files are not present, generating the docs will take a long time since Lean will have to compile all of mathlib from scratch.

jcommelin:
Which easily takes 2 hours.

readme.md
waldyrious: Sure, I can remove the link — just changing from "olean files" to "`.olean` files" already helps making things a little clearer, IMO. I could replace the link with https://leanprover-community.github.io/leanproject.html instead, but I don't think that page (in its current state) provides the information one would expect when clicking on the expression ".olean files" (i.e. additional details on what .olean files are).

That said, depending on the outcome of the conversation at https://github.com/leanprover-community/leanprover-community.github.io/pull/180, it could make sense to add the link here. I'd suggest holding this PR until that one is resolved, so we can decide what makes the most sense here.

PR #132, opened 2021-05-14
allow deploy action to work on PRs from external forks

robertylewis:
See https://github.com/leanprover-community/doc-gen/pull/115#issuecomment-779953415 , https://github.com/leanprover-community/doc-gen/pull/131

I think this test is unnecessary, since the action triggered is in leanprover-community/doc-gen:master and not in the PR. It should have access to the repo secrets.


gebner:
I have no idea if it will work, but I don't think this can do any harm.

PR #134, opened 2021-07-13
Upload the json export to github

eric-wieser:
I just wanted access to this file, and creating this PR was a means to obtain it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,3 +53,6 @@ jobs:
           github_repo: ${{ github.repository }}
           github_event: ${{ github.event_name }}
           github_ref: ${{ github.ref }}
+      - name: Upload export.json
+        uses: actions/upload-artifact@v2.2.4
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -53,3 +53,6 @@ jobs:
           github_repo: ${{ github.repository }}
           github_event: ${{ github.event_name }}
           github_ref: ${{ github.ref }}
+      - name: Upload export.json
+        uses: actions/upload-artifact@v2.2.4
+        path: export.json
```
eric-wieser, line 58:
```suggestion
        with:
          path: export.json
```

gebner:
Feel free to also deploy the export.json to the website if you'd like!

eric-wieser:
Oh, I wasn't actually planning to merge this in case it uses up some storage limit for build artifacts! Maybe we get nowhere close to that anyway.

PR #142, opened 2021-10-30
Output a gexf file for exploring the import graph

eric-wieser:
This is much much smaller than the export.json, but has plenty of information in to produce graphs from.

https://eric-wieser.github.io/doc-gen/import-graph.html makes use of this file. Given the dependencies of the graph, it might make sense not to host the actual visualizer within the doc-gen part of the mathlib docs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,8 +57,3 @@ jobs:
         uses: actions/upload-artifact@v2.2.4
         with:
           path: export.json
-
-      - name: Upload import.gexf
-        uses: actions/upload-artifact@v2.2.4
-        with:
-          path: mathlib_docs/docs/import.gexf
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -57,3 +57,8 @@ jobs:
         uses: actions/upload-artifact@v2.2.4
         with:
           path: export.json
+
+      - name: Upload import.gexf
+        uses: actions/upload-artifact@v2.2.4
+        with:
+          path: mathlib_docs/docs/import.gexf
```
eric-wieser, line 64:
```suggestion
```
This was just needed for me to get hold of a copy; in future it should end up uploaded alongside the `decl.bmp`.

print_docs.py
```diff
@@ -220,6 +220,9 @@ def of(cls, fname: str):
   def name(self):
     return '.'.join(self.parts)
 
+  def __str__(self):
+    return f'{self.proj}:{self.name}'  # primarily for networkx export
```
eric-wieser, line 224:
```suggestion
    return f'{self.project}:{self.name}'  # primarily for networkx export
```

bryangingechen (approved):
LGTM.

PR #144, opened 2021-12-11
Add canonical urls

eric-wieser:
I haven't tested this, and am not all that familiar with jinja. Let's see if this works...


eric-wieser:
#deploy

eric-wieser:
(where "success" means "now you must wait for github to actually publish to github-pages)

eric-wieser:
Seems to be working

PR #148, opened 2021-12-17
Convert to an installable python package

eric-wieser:
This isn't a full conversion, but should still work after this restructure.


gen_docs
```diff
@@ -36,4 +36,4 @@ bash "$source"scripts/mk_all.sh
 lean src/entrypoint.lean >export.json
 bash "$source"scripts/rm_all.sh
 
-python3 print_docs.py -r "$source" -w "$site_root" "${args[@]}"
+lean-doc-gen -r "$source" -w "$site_root" "${args[@]}"
```
gebner, line 39: Please make it possible to run doc-gen and gen_docs without installing the package.  `python -m lean_doc_gen` fails here.

gebner:
What is the story for the *.lean files here?  Are these installed as well?

> Since bors isn't enabled in this repository, if we do decide to merge this we should either squash merge through github, or I should manually do so locally.

Yeah sure, I can squash.

gen_docs
eric-wieser: Why do we care about supporting non-installed packages?

Note that even with the package installed, `python -m lean_doc_gen` fails here; but that's only because it picks up the non-installed version in the current directory. If I moved `lean_doc_gen` to `python_src/lean_doc_gen`, or just changed the working directory when running this script, then that wouldn't happen.

eric-wieser:
> What is the story for the *.lean files here? Are these installed as well?

No, not yet - but the python code doesn't use these at all yet.

gebner:
> No, not yet - but the python code doesn't use these at all yet.

Hmm, but is there anything you can do with the python code alone?  Related question: are the CSS/JS files copied as well?  I can't find any relevant changes in the PR.

PR #149, opened 2021-12-21
Make sorried definitions stand out in the docs

eric-wieser:
While this doesn't matter for mathlib, downstream projects may be more willing to allow partial formalizations.

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -139,6 +139,7 @@ meta structure decl_info :=
 (line : ℕ)
 (attributes : list string) -- not all attributes, we have a hardcoded list to check
 (noncomputable_reason : option _root_.name)
+(sorried : bool)
 (equations : list efmt)
 (kind : string) -- def, theorem, constant, axiom, structure, inductive
 (structure_fields : list (string × efmt)) -- name and type of fields of a constructor
@@ -170,7 +171,7 @@ meta instance {α} [has_coe α json] : has_coe (option α) json :=
 ⟨λ o, match o with | some x := ↑x | none := json.null end⟩
 
 meta def decl_info.to_json : decl_info → json
-| ⟨name, is_meta, args, type, doc_string, filename, line, attributes, nc_reason, equations, kind, structure_fields, constructors⟩ :=
+| ⟨name, is_meta, args, type, doc_string, filename, line, attributes, nc_reason, sorried, equations, kind, structure_fields, constructors⟩ :=
 json.object [
   ("name", to_string name),
   ("is_meta", is_meta),
@@ -181,6 +182,7 @@ json.object [
   ("line", line),
   ("attributes", json.of_string_list attributes),
   ("noncomputable_reason", nc_reason.map to_string),
+  ("sorried", sorried)
   ("equations", equations.map efmt.to_json),
   ("kind", kind),
   ("structure_fields", json.array $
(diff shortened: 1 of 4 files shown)


src/export_json.lean
```diff
@@ -139,6 +139,7 @@ meta structure decl_info :=
 (line : ℕ)
 (attributes : list string) -- not all attributes, we have a hardcoded list to check
 (noncomputable_reason : option _root_.name)
+(sorried : bool)
```
gebner, line 142: Please also add to the schema in the module docstring.

templates/decl.j2
```diff
@@ -54,5 +54,8 @@
     </details>
 {% endif %}
 
+{% if decl.sorried %}
+    <p> class="sorry_msg"><strong>Note:</strong> this declaration is incomplete and uses <code>sorry</code>.
```
gebner, line 58:
```suggestion
    <p> class="sorry_msg"><strong>Note:</strong> this declaration is incomplete and uses <code>sorry</code>. </p>
```
It doesn't matter if it's the last element before the `</div>`, but maybe somebody else will add another if below this.

style.css
```diff
@@ -454,6 +454,22 @@ main h2, main h3, main h4, main h5, main h6 {
...
 }
 
+.decl.sorried {
+    background-color: #fff0f0;
+}
+
+.decl .sorry_msg {
+    background-color: #ffdada;
+    /* this pushes it flush with the colored margin */
+    margin-left: -8px;
```
gebner, line 464: Do you have this deployed somewhere so that I can look at it?

templates/decl.j2
eric-wieser: This was just clumsy, I wouldn't knowingly skip a `<\p>`!

style.css
eric-wieser: Working on it, CI seems to be swallowing my error messages. Thanks for the speedy review.

templates/decl.j2
gebner: It's perfectly valid HTML5 though, the closing tag is implicit in some cases (see "tag omission" in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p).

src/export_json.lean
```diff
@@ -181,6 +183,7 @@ json.object [
   ("line", line),
   ("attributes", json.of_string_list attributes),
   ("noncomputable_reason", nc_reason.map to_string),
+  ("sorried", sorried)
```
eric-wieser, line 186:
```suggestion
  ("sorried", sorried),
```

eric-wieser:
https://pygae.github.io/lean-ga-docs/geometric_algebra/from_mathlib/mathoverflow.html should have a demo shortly

templates/decl.j2
```diff
@@ -54,5 +54,8 @@
     </details>
 {% endif %}
 
+{% if decl.sorried %}
+    <p> class="sorry_msg"><strong>Note:</strong> this declaration is incomplete and uses <code>sorry</code>. </p>
```
eric-wieser, line 58:
```suggestion
    <p class="sorry_msg"><strong>Note:</strong> this declaration is incomplete and uses <code>sorry</code>.</p>
```

eric-wieser:
Surprisingly it seems that `contains_sorry` only returns true for lemmas that use `sorry` directly, rather than transitively poisoning them. I think this is still better than nothing though, and maybe "depends on a sorried lemma" should be handled at the same time as "depends on an unusual axiom"

PR #151, opened 2022-01-12
Copy binder types from actual declaration, fixes display of or.inl

alexjbest:
Currently doc-gen prints `or.inl` as
```
inductive or (a b : Prop) :
Prop
inl : ∀ (a b : Prop), a → a ∨ b
inr : ∀ (a b : Prop), b → a ∨ b
```
but `#check or.inl` gives `or.inl : ∀ {a b : Prop}, a → a ∨ b` with implicit `a b`.
To fix this we copy the binder types from the actual declaration to the local constants used to pretty print the type

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -267,13 +267,18 @@ meta def mk_const_with_params (d : declaration) : expr :=
 let lvls := d.univ_params.map level.param in
 expr.const d.to_name lvls
 
+meta def update_binders_with : list expr → list binder → list expr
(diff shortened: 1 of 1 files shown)


alexjbest:
#deploy

gebner:
#deploy

alexjbest:
Ok, https://leanprover-community.github.io/mathlib_docs_demo/init/core.html#or looks better now, I checked a few other inductives and they seem fine too.

PR #153, opened 2022-01-28
feat: use STIX Two Math as a backup font

ericrbg:
One pitfall - the license is the [OFL license](https://github.com/stipub/stixfonts/blob/master/OFL.txt), which is free but I'm not sure where we have to put the attribution. Other than that I think this should be all good to go.


gebner:
#deploy

gebner:
> but I'm not sure where we have to put the attribution.

This is a very good question.  From what I can tell, we're using the unmodified font files here (which means we don't need to worry about the Reserved Font Names condition).

After reading the license / official FAQ, I believe we certainly need to include the license text in the doc-gen repository (because we distribute the font file).  I'd also copy the license file next to the font file on the website.

I don't think it is required to put attribution in the website itself.

ericrbg:
I think that's now been done - also, I checked and had a browse in a couple different browsers, and nothing seems to have broken from doing this, thankfully.

style.css
```diff
@@ -472,7 +476,7 @@ main h2, main h3, main h4, main h5, main h6 {
 
 .imports li, code, .decl_header, .attributes, .structure_field,
         .constructor, .instances li, .equation, #search_results div {
-    font-family: 'Source Code Pro', monospace;
+    font-family: 'Source Code Pro', monospace, 'STIX Two Math';
```
gebner, line 479:
```suggestion
    font-family: 'Source Code Pro', 'STIX Two Math', monospace;
```

For reproducibility.  `monospace` is the system-provided font.

ericrbg: I was worried that there would be some non-maths characters that would end up showing as non-monospace STIX Math characters, but this doesn't seem to be the case. I'll also make STIX the backup font for normal text, so that, e.g., ℤ prints the same inside and outside source code.

PR #155, opened 2022-01-28
Stripping whitespaces from user queries

arthurpaulino:
Also adding a tip (thanks Gabriel!) for computers without enough resources to generate the `html` directory locally.


gebner:
Might not be the ideal solution but it doesn't cost us anything and is better than what we have now.

PR #158, opened 2022-02-08
add dark theme

cuppajoeman:
Aside from the dark theme this commit adds a button which lets you toggle the color scheme which saves your setting between sessions.


eric-wieser:
#deploy

eric-wieser:
Hmm, [that crashed](https://github.com/leanprover-community/doc-gen/runs/5105797765?check_suite_focus=true)...

eric-wieser:
#deploy

eric-wieser:
@gebner, any idea what's causing deployment failures? Is this the "runner is too puny" problem that motivated self-hosted runners elsewhere?

ericrbg:
hmm, I doubt this is what it is, but you need to tell the main `.py` to copy stuff over for it to work, see my changes to `print_docs.py` [here](https://github.com/leanprover-community/doc-gen/pull/153/files)

gebner:
> any idea what's causing deployment failures? Is this the "runner is too puny" problem that motivated self-hosted runners elsewhere?

Yes, I only moved the regular CI over to the self-hosted runners but forgot about `#deploy` since that's a separate workflow.

#deploy

detect_color_scheme.js
```diff
@@ -0,0 +1,16 @@
...
+        var theme = "dark";
+    }
+} else if(window.matchMedia("(prefers-color-scheme: dark)").matches) {
+    // OS theme setting detected as dark
+    var theme = "dark";
+}
+
+//dark theme preferred, set document with a `data-theme` attribute
+if (theme=="dark") {
+     document.documentElement.setAttribute("data-theme", "dark");
```
gebner, line 15: This will only pick up the setting the first time the page is loaded.  If I switch the system setting from light to dark, the docs will still be light.

Maybe the easiest way would be to just integrate https://github.com/GoogleChromeLabs/dark-mode-toggle, which seems to handle this case quite robustly.

style.css
```diff
@@ -20,10 +29,49 @@ nav { line-height: normal; }
 
 :root {
     --header-height: 3em;
+    /* custom */
     --header-bg: #f8f8f8;
+    --body-bg: white;
+    --code-bg: #f3f3f3;
+    --text-color: black;
+    --text-alt-color: hsl(210, 100%, 30%);
```
gebner, line 37:
```suggestion
    --link-color: hsl(210, 100%, 30%);
```

style.css
```diff
@@ -20,10 +29,49 @@ nav { line-height: normal; }
...
     --fragment-offset: calc(var(--header-height) + 1em);
     --content-width: 55vw;
 }
+
+[data-theme="dark"] {
+    --header-bg: #181717;
+    --body-bg: #333;
+    --code-bg: #363333;
+    --text-color: #eee;
+    --text-alt-color: #58a6ff;
```
gebner, line 60: I've found the yellow a bit glaring in the screenshot you've posted.  How about something more subdued, like `#fad956` (copied from v8.dev)?

eric-wieser: Note that this is already changed, `#58a6ff` is not yellow

style.css
cuppajoeman: Changed, the variable name in the newest commit.

gebner:
#deploy

detect_color_scheme.js
cuppajoeman: My newest commit should fix this, but I haven't had a chance to test it.

gebner:
#deploy

detect_color_scheme.js
cuppajoeman: > This will only pick up the setting the first time the page is loaded. If I switch the system setting from light to dark, the docs will still be light.
> 
> Maybe the easiest way would be to just integrate https://github.com/GoogleChromeLabs/dark-mode-toggle, which seems to handle this case quite robustly.

Does the newest commit solve this for you?

theme_toggler.js
```diff
@@ -0,0 +1,27 @@
...
+        localStorage.setItem('theme', 'light');
+        document.documentElement.setAttribute('data-theme', 'light');
+    }    
+    toggled = !toggled;
+}
+
+toggleSwitch.addEventListener('click', switchTheme, false);
+
+// also check to see if the user changes their theme settings while the page is loaded.
+window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', event => {
```
eric-wieser, line 20: This should probably also either clear or override the local storage

cuppajoeman: Newest changes should fix this.

detect_color_scheme.js
```diff
@@ -0,0 +1,16 @@
...
+// If they have stored settings, use those. Otherwise use their OS's theme.
+if (storedTheme) {
+    theme = storedTheme
+} else if (OSThemeIsDark) {
+    theme = "dark"
+}
+
+// this changes the value of the css varaibles to make it dark
+if (theme=="dark") {
+     document.documentElement.setAttribute("data-theme", "dark");
```
eric-wieser, line 15: nit: indent is weird here
```suggestion
    document.documentElement.setAttribute("data-theme", "dark");
```

eric-wieser (commented):
There are a few leftover light colors, like:
https://github.com/leanprover-community/doc-gen/blob/8b81b1a117a3cfcdc4660a68e8ea423e39a78cf7/style.css#L146-L152

and

https://github.com/leanprover-community/doc-gen/blob/8b81b1a117a3cfcdc4660a68e8ea423e39a78cf7/style.css#L617-L618

and

https://github.com/leanprover-community/doc-gen/blob/8b81b1a117a3cfcdc4660a68e8ea423e39a78cf7/style.css#L629-L630

and

https://github.com/leanprover-community/doc-gen/blob/8b81b1a117a3cfcdc4660a68e8ea423e39a78cf7/style.css#L208

style.css
```diff
@@ -20,10 +29,49 @@ nav { line-height: normal; }
...
+    --axiom-and-constant-color-hsl-angle: 16;
+    --structure-and-inductive-color: #f0a202;
+    --structure-and-inductive-color-hsl-angle: 40;
+    --starting-percentage: 100;
+    
     --fragment-offset: calc(var(--header-height) + 1em);
     --content-width: 55vw;
 }
+
+[data-theme="dark"] {
```
eric-wieser, line 55:
```suggestion
[data-theme="dark"] {
    color-scheme: dark;

```
which changes the scrollbars to dark too

gebner:
#deploy

digama0:
There is a black text header at https://leanprover-community.github.io/mathlib_docs_demo/commands.html##explode%20/%20#explode_widget, and the red color of `Prop` in the code block is too dark.

gebner:
> Does the newest commit solve this for you?

Not really.  Changing the system style doesn't take effect immediately on the docs website.  Sometimes I also have to press the toggle twice before it takes effect.

Two other things:
 - The toggle button is not visible on portrait (less than 700px wide)
 - The hamburger button only has light colors (also for portrait mode).
 - Between ~700-800px window width, the toggle button overflows into a second line (because it is too wide).
 - The toggle button is inside the search form (that might get sent to google)

What I would suggest for the portrait styling issues is to 1) put the toggle button after the form, 2) remove the header-side-padding, and 3) put an empty <div></div> at the beginning of the header.  I would also change the button text to something shorter and more descriptive, that is, "dark" or "🌓" instead of "toggle theme".

cuppajoeman:
I've added some more changes today to add more variables with respect to Eric's post. 

I see that @gebner still has the issue with his operating system color scheme changes not affecting the site, to that end I don't really know why that `window.matchMedia` doesn't work, so if anyone wants to take a look at that please do.

As for the other styling issues that gebner mentioned, I can try to do that as well, but it probably wouldn't happen until the end of the day today or in the next day, so if anyone wants to take that on today then there wouldn't be any conflicts.

cuppajoeman:
> 



> > Does the newest commit solve this for you?
> 
> Not really. Changing the system style doesn't take effect immediately on the docs website. Sometimes I also have to press the toggle twice before it takes effect.
> 
> Two other things:
> 
>     * The toggle button is not visible on portrait (less than 700px wide)
> 
>     * The hamburger button only has light colors (also for portrait mode).
> 
>     * Between ~700-800px window width, the toggle button overflows into a second line (because it is too wide).
> 
>     * The toggle button is inside the search form (that might get sent to google)
> 
> 
> What I would suggest for the portrait styling issues is to 1) put the toggle button after the form, 2) remove the header-side-padding, and 3) put an empty `<div></div>` at the beginning of the header. I would also change the button text to something shorter and more descriptive, that is, "dark" or "first_quarter_moon" instead of "toggle theme".

I've added these changes

eric-wieser:
#deploy

robertylewis:
Some weird-looking spots, on a spotcheck:

image
(same on all the "general documentation" pages)

This appears everywhere except the index page, and clicking it does nothing:
image

robertylewis:
I would also prefer the toggle to be more subtle than a big button at the top. As someone who doesn't use dark themes, I may be blind to this. But it isn't common web design to put this kind of toggle so prominently on a page, is it?

cuppajoeman:
> Some weird-looking spots, on a spotcheck:
> 
> image (same on all the "general documentation" pages)
> 
> This appears everywhere except the index page, and clicking it does nothing: image

Thanks for catching these - I'll fix up these things tonight. 

As for your other comment - I'm not really sure whether it's good practice or not to put the button up there, but I've definitely seen websites which do that.

cuppajoeman:
When I load up the tactics page it looks like this: 
image

which is unlike what @robertylewis was reporting, so probably they just need to refresh the page to get the new css (ctrl-shift-r) usually works for that. Besides that I've updated the button position and removed the old toggle button, can we rebuild and see how it looks?

bryangingechen:
#deploy

cuppajoeman:
I think with the newest changes this seems almost good to go!

PR #160, opened 2022-03-01
Adjusted dark theme

eric-wieser:
This changes the toggle button to be hidden in the left sidebar, and adds a third "system" option.
The old behavior was to override the users choice whenever the system settings changes, but this is no longer necessary.

This also fixes the `sorried` colors in dark mode, to:

image

Finally, this tweaks some navbar styles so that its right-hand side isn't clipped on small screens, as the color scheme options now sit against that right edge.


eric-wieser:
#deploy

eric-wieser:
#deploy

gebner:
The color_scheme.js file seems to be missing,.

eric-wieser:
#deploy

gebner:
BTW, `#deploy` takes a bit longer these days since we need to run the documentation generation on our self-hosted runners now.

eric-wieser:
#deploy

eric-wieser:
#deploy

gebner:
#deploy

PR #161, opened 2022-03-03
Add opensearch support

eric-wieser:


eric-wieser:
#deploy

eric-wieser:
This seems not to work

eric-wieser:
I think Google only respects the meta tag if it finds it on the homepage at the root of the URL

gebner:
Yeah, I can add it in firefox but not in chromium.

And searching with spaces is not supported: https://leanprover-community.github.io/mathlib_docs_demo/find/nat+cast

PR #163, opened 2022-04-04
Show instances available for a given type

eric-wieser:
I have no idea what I'm doing with this meta code, so it's entirely plausible there is a much better way to do this.

The heuristics for choosing what to display are overly generous, mainly because I don't know how to work with binders effectively in meta code.
Ideally instances about `list.lex` wouldn't show up under the list of instances for list, but that would need something that notices when type arguments to typeclasses are used as indices to later type arguments to typeclasses.

Either way, it seems to be collecting moderately helpful information.

We might want a special page that lists instances on pi, Type, Prop, and Sort; this PR collects the data for instances on those types, but doesn't attempt to render it anywhere

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/doc-gen.23163.20-.20reverse.20instance.20lookup/near/278550929)

### DIFF
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -436,6 +436,8 @@ pure $ json.object [
 ]
 
 open lean.parser
+
+/-- Open all locales (to turn on notation, and also install some notation hacks). -/
 @[user_command]
 meta def open_all_locales (_ : interactive.parse (tk "open_all_locales")): lean.parser unit :=
 do m ← of_tactic localized_attr.get_cache,
@@ -443,13 +445,12 @@ do m ← of_tactic localized_attr.get_cache,
    cmds.mmap' $ λ m,
     when (¬ ∃ tok ∈ m.split_on '`', by exact
         (tok.length = 1 ∧ tok.front.is_alphanum) ∨ tok ∈ ["ε", "φ", "ψ", "W_", "σ", "ζ", "μ", "π"]) $
-    lean.parser.emit_code_here m <|> skip
+    lean.parser.emit_code_here m <|> skip,
+   -- HACK: print gadgets with less fluff
+   lean.parser.emit_code_here "notation x ` := `:10 y := opt_param x y",
+   -- HACKIER: print last component of name as string (because we can't do any better...)
+   lean.parser.emit_code_here "notation x ` . `:10 y := auto_param x (name.mk_string y _)"
 
 meta def main : io unit := do
 json ← run_tactic (trace_error "export_json failed:" mk_export_json),
 put_str json.unparse
-
--- HACK: print gadgets with less fluff
-notation x ` := `:10 y := opt_param x y
--- HACKIER: print last component of name as string (because we can't do any better...)
-notation x ` . `:10 y := auto_param x (name.mk_string y _)
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -340,10 +340,8 @@ do docs ← olean_doc_strings,
      (filename, json.array $ l.map write_module_doc_pair) :: rest
    end) [])
 
-#check @has_coe_to_sort.coe
-
 /--
-Three types of failure:
+The return type is to indicate three types of failure:
 * tactic failure (bug)
 * heuristic failure `exceptional.exception`
 * no name available `none`
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -343,9 +343,9 @@ do docs ← olean_doc_strings,
 /-- Extract `[foo, bar]` from `has_pow foo bar`.
 
 The result is a map containing the type names as keys.  -/
-meta def find_instance_types (e : expr) : tactic (rb_map string unit) :=
+meta def find_instance_types (e : expr) : tactic (rb_set string) :=
 list.mfoldl (λ m e, do
-  e ← dunfold [`decidable_rel, `decidable_pred] e {fail_if_unchanged := ff},
+  -- e ← dunfold [`decidable_rel, `decidable_pred] e {fail_if_unchanged := ff},
   e ← whnf e transparency.reducible,
   t ← infer_type e,
   -- only look at arguments which are types or propositions
@@ -364,8 +364,8 @@ list.mfoldl (λ m e, do
   | expr.app _ _             := fail ("app, not a constant: " ++ to_string e)
   | expr.elet _ _ _ _        := fail ("elet, not a constant: " ++ to_string e)
   end | pure m,
-  pure (m.insert s ())
-) mk_rb_map e.get_app_args 
+  pure (m.insert s)
+) mk_rb_set e.get_app_args 
 
 meta def get_instances : tactic (rb_lmap string string × rb_lmap string string) :=
 attribute.get_instances `instance >>= list.mfoldl
@@ -377,7 +377,7 @@ attribute.get_instances `instance >>= list.mfoldl
         fail ("not a constant: " ++ to_string e),
       types ← find_instance_types e,
       let new_rev_map :=
-        types.fold rev_map (λ arg _ rev_map,  rev_map.insert arg inst_nm.to_string),
+        types.fold rev_map (λ arg rev_map, rev_map.insert arg inst_nm.to_string),
       return $ (map.insert class_nm.to_string inst_nm.to_string, new_rev_map))
   (mk_rb_map, mk_rb_map)
 
--- a/src/export_json.lean
+++ b/src/export_json.lean
@@ -340,30 +340,52 @@ do docs ← olean_doc_strings,
      (filename, json.array $ l.map write_module_doc_pair) :: rest
    end) [])
 
+#check @has_coe_to_sort.coe
+
+/--
+Three types of failure:
+* tactic failure (bug)
+* heuristic failure `exceptional.exception`
+* no name available `none`
+-/
+meta def extract_name : expr → tactic (exceptional (option string))
+| e := do
+  e ← whnf e transparency.reducible,
+  -- t ← infer_type e,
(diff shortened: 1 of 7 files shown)


eric-wieser:
#deploy

ericrbg:
i'd be very happy to see this!

gebner:
#deploy

gebner:
Deploy failed because this branch doesn't have 330e01df63a71a645abe664727584fb3932326d8 yet.

src/export_json.lean
```diff
@@ -340,20 +340,52 @@ do docs ← olean_doc_strings,
      (filename, json.array $ l.map write_module_doc_pair) :: rest
    end) [])
 
-meta def get_instances : tactic (rb_lmap string string) :=
+/-- Extract `[foo, bar]` from `has_pow foo bar`.
+
+The result is a map containing the type names as keys.  -/
+meta def find_instance_types (e : expr) : tactic (rb_map string unit) :=
```
gebner, line 346:
```suggestion
meta def find_instance_types (e : expr) : tactic (rb_set string) :=
```

src/export_json.lean
```diff
@@ -340,20 +340,52 @@ do docs ← olean_doc_strings,
      (filename, json.array $ l.map write_module_doc_pair) :: rest
    end) [])
 
-meta def get_instances : tactic (rb_lmap string string) :=
+/-- Extract `[foo, bar]` from `has_pow foo bar`.
```
gebner, line 343: This function will also extract the implicit arguments from the typeclass, which results in all kinds of spurious instances on real.

src/export_json.lean
```diff
@@ -340,20 +340,52 @@ do docs ← olean_doc_strings,
...
+
+The result is a map containing the type names as keys.  -/
+meta def find_instance_types (e : expr) : tactic (rb_map string unit) :=
+list.mfoldl (λ m e, do
+  e ← dunfold [`decidable_rel, `decidable_pred] e {fail_if_unchanged := ff},
+  e ← whnf e transparency.reducible,
+  t ← infer_type e,
+  -- only look at arguments which are types or propositions
+  expr.sort l ← whnf t transparency.reducible | pure m,
```
gebner, line 352: I'm not sure if the restriction is really necessary.

src/export_json.lean
```diff
@@ -340,20 +340,52 @@ do docs ← olean_doc_strings,
      (filename, json.array $ l.map write_module_doc_pair) :: rest
    end) [])
 
-meta def get_instances : tactic (rb_lmap string string) :=
+/-- Extract `[foo, bar]` from `has_pow foo bar`.
+
+The result is a map containing the type names as keys.  -/
+meta def find_instance_types (e : expr) : tactic (rb_map string unit) :=
+list.mfoldl (λ m e, do
+  e ← dunfold [`decidable_rel, `decidable_pred] e {fail_if_unchanged := ff},
```
gebner, line 348: How does this even happen?  `e` is an *argument* after all.

src/export_json.lean
```diff
@@ -340,20 +340,52 @@ do docs ← olean_doc_strings,
...
+  | expr.sort level.zero     := pure (some "Prop")
+  | expr.sort (level.succ l) := pure (some "Type")
+  | expr.sort l              := pure (some "Sort")
+  | expr.local_const _ _ _ _ := pure none
+  | expr.macro _ _           := pure none -- TODO: unfold macros?
+  | expr.lam _ _ _ _         := fail ("lam, not a constant: " ++ to_string e)
+  | expr.var i               := fail ("var, not a constant: " ++ to_string e)
+  | expr.mvar _ _ _          := fail ("mvar, not a constant: " ++ to_string e)
+  | expr.app _ _             := fail ("app, not a constant: " ++ to_string e)
+  | expr.elet _ _ _ _        := fail ("elet, not a constant: " ++ to_string e)
```
gebner, line 365: This bails out if any argument is a lambda, but we probably just want to ignore the lambda arguments.

templates/decl.j2
```diff
@@ -54,6 +54,17 @@
     </details>
 {% endif %}
 
+{% if decl.name in instances_for %}
+    <details class="instances-for">
+        <summary>Available instances</summary>
```
gebner, line 59: I found the terminology a bit confusing.  We have "instances" and "available instances", but I couldn't tell you which is which.  maybe we should look at how the haskell documentation calls this.  Maybe "instances of this type class" and "implements the following type classes"?

src/export_json.lean
eric-wieser: I think this was motivated mainly by:

* Not having a good idea how to word the docs for the non-type cases
* Trying to limit the growth of the binary size of the doc

src/export_json.lean
gebner: How about this?
```lean
meta def get_explicit_args (e : expr) : tactic (list expr) :=
fold_explicit_args e [] (λ xs x, pure $ xs ++ [x])
```

src/export_json.lean
gebner: This is an example where you have a lambda as argument: https://leanprover-community.github.io/mathlib_docs_demo/algebra/graded_monoid.html#set_like.ghas_one

But I guess we skip this argument because it's not a type.

> bailing out

I didn't read the code at first, but apparently the fail is fatal.  (I blindly assumed there was a `<|> pure []` or something afterwards.)  I'd really like to avoid such fatal errors.  I'm already dreading the email from github actions that says "doc-gen failed, elet, not a constant: ...".

src/export_json.lean
eric-wieser: I no longer remember, I think this must be vestigial from an earlier version

src/export_json.lean
eric-wieser: Excellent, I didn't know about `fold_explicit_args`

templates/decl.j2
eric-wieser: Haskell seems to use "instances" for both versions:

* https://hackage.haskell.org/package/base-4.16.1.0/docs/Foreign-C-Types.html#t:CBool (instances on a type)
* https://hackage.haskell.org/package/base-4.16.1.0/docs/Prelude.html#t:Integral (instances of a typeclass)

The reality is that probably we never show both these headings anyway

src/export_json.lean
```diff
@@ -340,20 +340,50 @@ do docs ← olean_doc_strings,
...
+meta def get_instances : tactic (rb_lmap string string × rb_lmap string string) :=
 attribute.get_instances `instance >>= list.mfoldl
-  (λ map inst_nm,
+  (λ ⟨map, rev_map⟩ inst_nm,
    do ty ← mk_const inst_nm >>= infer_type,
-      (_, e) ← open_pis_whnf ty transparency.reducible,
+      (binders, e) ← open_pis_whnf ty transparency.reducible,
       e ← whnf e transparency.reducible,
       expr.const class_nm _ ← pure e.get_app_fn |
         fail ("not a constant: " ++ to_string e),
```
eric-wieser, line 375: @gebner, note that we already had a hard failure here anyway.

In my opinion getting the github email is a good thing; upon receipt, we change the hard failure to a TODO comment, but at least then we know what the problem is.

test/instances.lean
```diff
@@ -0,0 +1,23 @@
+/-
+A short example to test the behavior of `get_instances`.
+To actually test this, paste this code above the `HACK` lines in `export_json.lean`, as after those
+lines notation becomes badly broken.
+-/
```
eric-wieser, line 5: This test isn't actually executed by CI due to the problem that `export_json` poisons the notation, but a test that can be manually executed is better than no test at all.

src/export_json.lean
gebner: > In my opinion getting the github email is a good thing

The problem is that I'm personally getting spammed by these emails every few hours.  Until I either fix it or add an issue (and continue being spammed).  And in the meanwhile the docs are not being updated.

If we can change it so that you get the email and are responsible for fixing the issues, then I'd be happy with adding such easily breaking code.

As far as I'm concerned, the code in this function is purely a heuristic anyhow.  The most likely failure scenario is that someone on Zulip complains about an being included that shouldn't be (see e.g. implicit arguments) or not being included while it should be (see e.g. coercions).  This check won't help with that at all.

test/instances.lean
gebner: We could move the hacks to entrypoint.lean

templates/decl.j2
gebner: > The reality is that probably we never show both these headings anyway

Actually we do: https://leanprover-community.github.io/mathlib_docs_demo/algebra/group/defs.html#group

eric-wieser: Oh, nice find.

eric-wieser:
#deploy

src/export_json.lean
eric-wieser: I've gone ahead and made both failures report to stderr instead.

templates/decl.j2
eric-wieser: Let me know if the new logic looks better

src/export_json.lean
eric-wieser: I added some handling of lambdas

src/export_json.lean
eric-wieser: I've removed this restriction, I realized that I wanted instances to appear for `lp`.

eric-wieser:
#deploy

src/export_json.lean
gebner: Thanks!

gebner:
Thanks!

PR #164, opened 2022-04-29
Add support for gitpod development

eric-wieser:
This merges a typical lean gitpod setup with the python setup that gitpod autogenerated for this repo


gebner:
You might also want to update the mathlib revision.  I don't think doc-gen builds out of the box atm.

PR #166, opened 2022-05-11
Fix plaintext summary

eric-wieser:
This avoids getting nonsense like `quaternionalgebra` in the `description` field instead of `quaternion_algebra`.
This was caused by the "remove italics" code misfiring.

Depends on #165

### DIFF
--- a/mistletoe_renderer.py
+++ b/mistletoe_renderer.py
@@ -9,7 +9,7 @@ Extra features include:
 """
 import re
 
-from mistletoe import Document, HTMLRenderer, span_token
+from mistletoe import Document, HTMLRenderer, BaseRenderer, span_token
 from pygments import highlight
 from pygments.lexers import get_lexer_by_name as get_lexer
 from pygments.formatters.html import HtmlFormatter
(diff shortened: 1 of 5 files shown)


test/test_python.py
```diff
@@ -0,0 +1,40 @@
...
+    * `quaternion.conj` : conjugate of a quaternion;
+
+    We also define the following algebraic structures on `ℍ[R]`:
+
+    * `ring ℍ[R, a, b]` and `algebra R ℍ[R, a, b]` : for any commutative ring `R`;
+    * `ring ℍ[R]` and `algebra R ℍ[R]` : for any commutative ring `R`;
+    * `domain ℍ[R]` : for a linear ordered commutative ring `R`;
+    * `division_algebra ℍ[R]` : for a linear ordered field `R`.""")
+    expected = (
+        "Quaternions: In this file we define quaternions `ℍ[R]` over a commutative ring `R`, and "
```
gebner, line 31:
```suggestion
        "Quaternions: In this file we define quaternions ℍ[R] over a commutative ring R, and "
```

Would it make sense to strip the backticks as well?

gebner (commented):
LGTM, feel free to merge if you're satisfied.  This is certainly a nicer approach then the regex massaging we have right now.

mistletoe_renderer.py
```diff
@@ -114,3 +114,71 @@ def render_raw_url(self, token):
...
+    render_emphasis = BaseRenderer.render_inner
+
+    def render_inline_code(self, token):
+        return f"`{self.render_inner(token)}`"
+
+    def render_raw_text(self, token):
+        return token.content
+
+    render_strikethrough = BaseRenderer.render_inner
+    render_strikethrough = BaseRenderer.render_inner
```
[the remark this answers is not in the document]
eric-wieser: No, this is an error

test/test_python.py
eric-wieser: I'd prefer to keep them, they make it easier to separate words from lean-nonsense

eric-wieser:
Ugh, it turns out that `mistletoe` is not rentrant, so we can't have more than one live renderer at the same time.

https://github.com/executablebooks/mistletoe-ebp fixed this, but they never upstreamed their fix and abandoned it.

eric-wieser:
#deploy

PR #168, opened 2022-07-21
display automatic translations to informal math

robertylewis:
Thanks to work by @zhangir-azerbayev , we have a dataset of every declaration in (a recent version of) mathlib translated to natural language. People from OpenAI are interested in improving this; in principle we can regenerate these translations periodically, and update the model with user responses from the displayed translations. 

There is a static json file containing the translations. It also stores the theorem name and statement that were used to generate the text. doc-gen checks that the statement hasn't changed before printing the translation, so even if we go months between updates, the worst we'll see is a lot of theorems with no translation.

All of this is WIP:
* We need a doc page on the community website explaining what's going on here.
* I did not try hard at all on the display stylilng. Obviously this needs to be improved. Right now I put it in a `<details>` menu, expanded by default to make browsing this demo easier. I think they should be collapsed by default in production, at least until the translations improve.
* The response buttons don't go anywhere. We should have them pop up a window that allows the user, optionally, to correct the statement and submit. They can target a Google form or any other database.

### DIFF
--- a/nav.js
+++ b/nav.js
@@ -250,26 +250,48 @@ for (const elem of document.getElementsByClassName('gh_link')) {
 // Informal Statement feedback form handler
 // -------
 
-window.addEventListener('load', event => {
-  const forms = document.querySelectorAll('form.informal_statement_form')
-  for (const form of forms) {
-    const url = new URL(form.getAttribute('action'))
-    form.addEventListener('submit', event => {
+window.addEventListener('load', _ => {
+  for (const translationDiv of document.querySelectorAll('.translation_qs')) {
+    const declName = translationDiv.getAttribute('data-decl')
+    if (!declName) {
+      console.error('no data-decl on translation_qs')
+    }
+    const feedbackForm = translationDiv.querySelector('.informal_statement_feedback')
+    const editForm = translationDiv.querySelector('.informal_statement_edit')
+    const ta = editForm.querySelector('textarea')
+    const url = new URL(feedbackForm.getAttribute('action'));
+    url.searchParams.set('decl', declName)
+    url.searchParams.set('statement', ta.value)
+    feedbackForm.addEventListener('submit', async event => {
       event.preventDefault()
-      const name = event.submitter.getAttribute('name')
-      const value = event.submitter.getAttribute('value')
-      url.searchParams.set(name, value)
-      form.textContent = "Sending..."
-      fetch(url, {method: 'POST'}).then(response => {
-        if (response.ok) {
-          form.textContent = "Thanks for your feedback!"
-          // [todo] in future we can get server to return stats on how many other people found this correct etc.
-        } else {
-          form.textContent = "There was a server error."
+      try {
+        const name = event.submitter.getAttribute('name')
+        const value = event.submitter.getAttribute('value')
+        url.searchParams.set(name, value)
+        if (value === 'no') {
+          feedbackForm.textContent = "Please correct the text below:"
+          editForm.removeAttribute('style')
+          const edit = await new Promise((resolve, reject) => {
+            editForm.addEventListener('submit', event => {
+              event.preventDefault()
+              resolve(ta.value)
+            })
+          });
+          url.searchParams.set('edit', edit)
+          editForm.remove()
         }
-      }).catch(err => {
-        form.textContent = `Error: ${err.message}`
-      })
+        feedbackForm.textContent = "Sending..."
+
+        const response = await fetch(url, { method: 'POST' })
+        if (response.ok) {
+          feedbackForm.textContent = "Thanks for your feedback!"
+        } else {
+          const txt = await response.text()
+          throw new Error(`Bad response: ${txt}`)
+        }
+      } catch (err) {
+        feedbackForm.textContent = `Error: ${err.message}`
+      }
     })
   }
 })
--- a/nav.js
+++ b/nav.js
@@ -259,6 +259,7 @@ async function fetchGood(...args) {
   }
 }
 
+/** Handler for clicking feedback buttons in informal statements. */
 window.addEventListener('load', _ => {
   for (const translationDiv of document.querySelectorAll('.translation_qs')) {
     const declName = translationDiv.getAttribute('data-decl')
@@ -299,4 +300,31 @@ window.addEventListener('load', _ => {
       }
     })
   }
+})
+
+const INFORMAL_OPEN_ID = 'informal_statement_open';
+function updateInformalOpen(state) {
+  if (state !== undefined) {
+    localStorage.setItem(INFORMAL_OPEN_ID, state);
+  } else {
+    state = localStorage.getItem(INFORMAL_OPEN_ID) ?? true;
+  }
+  const details = document.querySelectorAll('.informal_statement_details');
+  for (const detail of details) {
+    if (state) {
+      detail.setAttribute('open', '');
+    } else {
+      detail.removeAttribute('open');
+    }
+  }
+}
+
+window.addEventListener('load', _ => {
+  updateInformalOpen();
+  const checkbox = document.querySelector('#informal-open');
+  checkbox.checked = localStorage.getItem(INFORMAL_OPEN_ID) ?? true
+  checkbox.addEventListener('change', e => {
+    const value = checkbox.checked
+    updateInformalOpen(value)
+  })
 })
--- a/nav.js
+++ b/nav.js
@@ -304,14 +304,14 @@ window.addEventListener('load', _ => {
 
 const INFORMAL_OPEN_ID = 'informal_statement_open';
 function getInformalOpen() {
-  let x = localStorage.getItem(INFORMAL_OPEN_ID) ?? 'true'
-  return x === 'true' ? true : false
+  let x = localStorage.getItem(INFORMAL_OPEN_ID) ?? 'true';
+  return x === 'true';
 }
 function updateInformalOpen(state) {
   if (state !== undefined) {
     localStorage.setItem(INFORMAL_OPEN_ID, state);
   } else {
-    state = getInformalOpen()
+    state = getInformalOpen();
   }
   const details = document.querySelectorAll('.informal_statement_details');
   for (const detail of details) {
--- a/nav.js
+++ b/nav.js
@@ -250,6 +250,15 @@ for (const elem of document.getElementsByClassName('gh_link')) {
 // Informal Statement feedback form handler
 // -------
 
+/** Same as `fetch` but throws an error if it's a bad response. */
+async function fetchGood(...args) {
+  const response = await fetch(...args);
+  if (!response.ok)  {
+    const txt = await response.text()
+    throw new Error(`Bad response: ${txt}`)
+  }
+}
+
 window.addEventListener('load', _ => {
   for (const translationDiv of document.querySelectorAll('.translation_qs')) {
     const declName = translationDiv.getAttribute('data-decl')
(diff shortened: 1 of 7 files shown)


robertylewis:
#deploy

zhangir-azerbayev:
@EdAyers Would you be interested in helping us with UI and databases?

style.css
```diff
@@ -697,3 +697,39 @@ a:hover {
...
+    color: #fff;
+    text-align: center;
+    padding: 5px 0;
+    border-radius: 6px;   
+    
+    /* Position the tooltip text - see examples below! */
+    top: -5px;
+    left: 105%;
+
+}
```
eric-wieser, line 722: I assume the `explain` css class this refers to was used in a previous version of the HTML / in an in-development version?

robertylewis: Yes, exactly.

robertylewis, line 722:
```suggestion
```

style.css
```diff
@@ -697,3 +697,39 @@ a:hover {
...
+
+.informal_statement .translation_qs {
+    color:gray;
+}
+  
+/* Show the tooltip text when you mouse over the tooltip container */
+.informal_statement .explain:hover .tooltiptext {
+    visibility: visible;
+}
```
robertylewis, line 735:
```suggestion
```

robertylewis:
When the informal output includes a `<` in math mode, it seems to be picked up as HTML before being picked up by MathJax: https://leanprover-community.github.io/mathlib_docs_demo/data/nat/basic.html#nat.one_add_le_iff (notice that the full statement is in the source code).

I thought MathJax was smart enough to do this, it handles `<` in doc strings perfectly fine for instance. Do we need to manually escape this character?

robertylewis:
@EdAyers if you're interested in helping with the UI, I have an approximate setup for the "data collection" part that needs to be hooked into the interface. We can chat.

templates/decl.j2
```diff
@@ -32,6 +32,29 @@
...
 {{ decl.doc_string | convert_markdown }}
 
+{% if decl.informal_statement %}
+
+<details open>
+    <summary>Informal translation</summary>
+
+    <div class="informal_statement">
+        <div class="statement">
+            {{ decl.informal_statement }}
```
bryangingechen, line 42:
```suggestion
            {{ decl.informal_statement | convert_markdown }}
```
The escaping code is in `convert_markdown` so this might fix things? Not tested though...

robertylewis:
#deploy

eric-wieser:
> I thought MathJax was smart enough to do this, it handles `<` in doc strings perfectly fine for instance. Do we need to manually escape this character?

Do you have a docstring in mind that proves this?

robertylewis:
> > I thought MathJax was smart enough to do this, it handles `<` in doc strings perfectly fine for instance.
> 
> Do you have a docstring in mind that proves this?

https://leanprover-community.github.io/mathlib_docs_demo/model_theory/order.html#first_order.language.sentence.densely_ordered is an example.

robertylewis:
Even weirder, https://leanprover-community.github.io/mathlib_docs_demo/model_theory/order.html#first_order.language.term.realize_lt has both a working and broken occurence of `<`!

robertylewis:
I thought I just posted this (eaten by GitHub somehow?), but, the problem is `<` followed by a non-space token. In principle this is an issue in normal doc strings as well as the automatic ones.

This is nontrivial but possible to replace during processing. This *must* have come up in relation to mathjax before, but I'm not turning up anything quickly on Google somehow.

eric-wieser:
This is not a mathjax problem, but a problem in our markdown parser. Mathjax cannot be reasonably expected to parse `<`, because by the time it runs the browser has already tried to parse it and made a mess.

eric-wieser:
I was planning to switch out our parser for markdown-it, but am waiting on an upstream patch...

robertylewis:
#deploy

templates/decl.j2
```diff
@@ -32,6 +32,35 @@
...
 
+{% if decl.informal_statement %}
+
+<details open>
+    <summary>Informal translation</summary>
+
+    <div class="informal_statement">
+        <div class="statement">
+            {{ decl.informal_statement | convert_markdown }}
+        </div>
```
eric-wieser, line 43: I'm worried the mathjax rendering is going to have quite a large impact on page load times. Could we maybe:

* Defer mathjax rendering of `.informal_statement` until the `details` is expanded
* Put a "expand informal math" setting in the sidebar next to the color scheme setting, so that users can opt to collapse everything by default and get the associated performance boost.

robertylewis: I think the `details` should be collapsed by default, not for performance but because the translations are misleadingly sketchy:
> Right now I put it in a `<details>` menu, expanded by default to make browsing this demo easier. I think they should be collapsed by default in production, at least until the translations improve.

But a toggle also sounds like a good idea.

Do you know how to defer the rendering?

robertylewis:
[`6c05dba`](https://github.com/leanprover-community/doc-gen/pull/168/commits/6c05dba9f69da42cc6e3a85385c71ff1f24d2977) is clearly a hack, but I think it should work until the markdown/mathjax interaction is fixed. (Hypothesis: the informal text will never have a `<` anywhere it isn't safe to insert a space after.)

robertylewis:
#deploy

robertylewis:
#deploy

EdAyers:
#deploy

robertylewis:
(the runners are backed up, the deploy may take a bit)

EdAyers:
I reckon this is ready to merge once @zhangir-azerbayev has fixed truncated translations?

templates/decl.j2
eric-wieser: There's a (possibly outdated) question about deferring here: https://stackoverflow.com/a/27952213/102441

eric-wieser: [This page in the documentation](https://docs.mathjax.org/en/latest/web/typeset.html#handling-asynchronous-typesetting) also seems relevant.

Enabling [Lazy Typesetting](https://docs.mathjax.org/en/latest/output/lazy.html#lazy-typesetting-1) would also be a reasonable option.

robertylewis:
> I reckon this is ready to merge once @zhangir-azerbayev has fixed truncated translations?

I've seen very few truncated translations, and regenerating the translations is costly. I think it's fine to wait to fix this until we're ready for a "next round." But there are still a few more to-dos:

* Eric's concerns above about delaying mathjax processing
* A toggle to collapse/expand all translation boxes
* A landing page on the website describing this project -- it's a broken link right now

nav.js
```diff
@@ -246,3 +246,89 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+        feedbackForm.textContent = `Error: ${err.message}`
+      }
+    })
+  }
+})
+
+const INFORMAL_OPEN_ID = 'informal_statement_open';
+function getInformalOpen() {
+  let x = localStorage.getItem(INFORMAL_OPEN_ID) ?? 'true'
+  return x === 'true' ? true : false
```
eric-wieser, line 308: Looks like we're fairly strict about semicolons elsewhere in this file, so
```suggestion
  let x = localStorage.getItem(INFORMAL_OPEN_ID) ?? 'true';
  return x === 'true' ? true : false;
```

nav.js
```diff
@@ -246,3 +246,89 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+}
+function updateInformalOpen(state) {
+  if (state !== undefined) {
+    localStorage.setItem(INFORMAL_OPEN_ID, state);
+  } else {
+    state = getInformalOpen()
```
eric-wieser, line 314:
```suggestion
    state = getInformalOpen();
```

nav.js
```diff
@@ -246,3 +246,89 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+  }
+  const details = document.querySelectorAll('.informal_statement_details');
+  for (const detail of details) {
+    if (state) {
+      detail.setAttribute('open', '');
+    } else {
+      detail.removeAttribute('open');
+    }
```
eric-wieser, line 322: Does `detail.open = state` work?

nav.js
```diff
@@ -246,3 +246,91 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+/** Handler for clicking feedback buttons in informal statements. */
+window.addEventListener('load', _ => {
+  for (const translationDiv of document.querySelectorAll('.translation_qs')) {
+    const declName = translationDiv.getAttribute('data-decl')
+    if (!declName) {
+      console.error('no data-decl on translation_qs')
+    }
+    const feedbackForm = translationDiv.querySelector('.informal_statement_feedback')
+    const editForm = translationDiv.querySelector('.informal_statement_edit')
+    const ta = editForm.querySelector('textarea')
```
eric-wieser, line 271:
```suggestion
    const declName = translationDiv.getAttribute('data-decl');
    if (!declName) {
      console.error('no data-decl on translation_qs');
    }
    const feedbackForm = translationDiv.querySelector('.informal_statement_feedback');
    const editForm = translationDiv.querySelector('.informal_statement_edit');
    const ta = editForm.querySelector('textarea');
```

nav.js
```diff
@@ -246,3 +246,91 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+          editForm.removeAttribute('style')
+          const edit = await new Promise((resolve, reject) => {
+            editForm.addEventListener('submit', event => {
+              event.preventDefault()
+              resolve(ta.value)
+            })
+          });
+          url.searchParams.delete('rate') // don't double-count the rating.
+          url.searchParams.set('edit', edit)
+          editForm.remove()
```
eric-wieser, line 293:
```suggestion
    url.searchParams.set('decl', declName);
    url.searchParams.set('statement', ta.value);
    feedbackForm.addEventListener('submit', async event => {
      event.preventDefault();
      try {
        const value = event.submitter.getAttribute('value');
        url.searchParams.set('rate', value);
        feedbackForm.textContent = "Sending...";
        await fetchGood(url, { method: 'POST' });
        if (value === 'no') {
          feedbackForm.textContent = "Thanks for your feedback! Optionally, please help us out by submitting a corrected statement: ";
          editForm.removeAttribute('style');
          const edit = await new Promise((resolve, reject) => {
            editForm.addEventListener('submit', event => {
              event.preventDefault();
              resolve(ta.value);
            });
          });
          url.searchParams.delete('rate'); // don't double-count the rating.
          url.searchParams.set('edit', edit);
          editForm.remove();
```

nav.js
```diff
@@ -246,3 +246,91 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+
+window.addEventListener('load', _ => {
+  updateInformalOpen();
+  const checkbox = document.querySelector('#informal-open');
+  checkbox.checked = getInformalOpen();
+  checkbox.addEventListener('change', e => {
+    const value = checkbox.checked;
+    updateInformalOpen(value);
+  });
+})
```
eric-wieser, line 336:
```suggestion
});

```

nav.js
```diff
@@ -246,3 +246,91 @@ for (const elem of document.getElementsByClassName('gh_link')) {
...
+const INFORMAL_OPEN_ID = 'informal_statement_open';
+
+function getInformalOpen() {
+  const item = localStorage.getItem(INFORMAL_OPEN_ID)
+  if (!item) {
+    return false // default is closed
+  } else {
+    return JSON.parse(item)
+  }
+}
```
eric-wieser, line 314:
```suggestion
function getInformalOpen() {
  const item = localStorage.getItem(INFORMAL_OPEN_ID);
  if (!item) {
    return false; // default is closed
  } else {
    return JSON.parse(item);
  }
}
```

robertylewis:
#deploy

robertylewis:
#deploy

robertylewis:
I think all of the to-dos are taken care of now (thanks @EdAyers !). I enabled lazy typesetting in mathjax, and confirmed it only processes when the display is unrolled.

PatrickMassot:
I think the warning should be more visible than using a light-grey color.

eric-wieser:
#deploy

(just so that we have something interesting at the docs_demo page)

PR #172, opened 2022-11-02
feat: add a new page with very basic documentation for `Pi`/`Sort`/`Type`/`Prop`

eric-wieser:
The actual text content is pretty sloppy here, the main aim is to make the instance lists available.

### DIFF
--- a/templates/foundational_types.j2
+++ b/templates/foundational_types.j2
@@ -89,4 +89,6 @@ Lean also permits ASCII-only spellings of the three variants:
 
 {% block internal_nav %}
 <h3><a href="#top">{{ self.title() }}</a></h3>
+
+{{ body.toc_html }}
 {% endblock %}
--- a/templates/foundational_types.j2
+++ b/templates/foundational_types.j2
@@ -1,0 +1,82 @@
+{% extends "base.j2" %}
+{% block title %}foundational types{% endblock %}
+
+{% block content %}
+
+{% filter convert_markdown %}
+Some of Lean's types are so primitive that they are available even in the `prelude`, and are not
+defined in any lean source files. This page provides some basic documentation on those types.
+
+
+For a more in depth explanation of Lean's type theory, refer to
+[TPiL](https://leanprover.github.io/theorem_proving_in_lean/dependent_type_theory.html).
+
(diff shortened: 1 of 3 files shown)


templates/foundational_types.j2
```diff
@@ -0,0 +1,82 @@
...
+
+{% block content %}
+
+{% filter convert_markdown %}
+Some of Lean's types are so primitive that they are available even in the `prelude`, and are not
+defined in any lean source files. This page provides some basic documentation on those types.
+
+
+For a more in depth explanation of Lean's type theory, refer to
+[TPiL](https://leanprover.github.io/theorem_proving_in_lean/dependent_type_theory.html).
```
kmill, line 12: A pass at wordsmithery:

```suggestion
Some of Lean's types are not defined in any Lean source files (even the `prelude`) since they come from its foundational type theory. This page provides basic documentation for these types.

For a more in-depth explanation of Lean's type theory, refer to
[TPiL](https://leanprover.github.io/theorem_proving_in_lean/dependent_type_theory.html).
```

templates/foundational_types.j2
```diff
@@ -0,0 +1,82 @@
...
+    <ul>
+        {% for inst in instances_for['Prop'] %}
+            <li>{{ inst | linkify }}</li>
+        {% endfor %}
+    </ul>
+</details>
+
+
+{% filter convert_markdown %}
+## Pi types (`Π a : α, β a` or `Pi a : α, β a` or `α → β`)
```
kmill, line 63: Should we include forall here?

templates/foundational_types.j2
```diff
@@ -0,0 +1,82 @@
...
+
+Pi types are the type of dependent functions. Note that function types are simply a special case
+of this.
```
kmill, line 66:
```suggestion
The type of dependent functions is known as a pi type. Non-dependent functions are a special case.
```

eric-wieser:
#deploy

eric-wieser:
#deploy

gebner:
That's an amazing addition!  At some point we should probably improve the instance lists a bit (maybe at least sort them, show the type, etc.).

PR #178, opened 2023-08-29
Make it clear this is for mathlib 3

eric-wieser:


eric-wieser:
We should wait for CI to pass before merging

PR #180, opened 2026-05-07
feat: add mathlib3 deprecation banner and point canonical URLs at mathlib4

kim-em:
This PR adds a fixed red banner across the top of every mathlib3 doc page, declaring that mathlib3 is deprecated and linking to the mathlib4 docs, and sets the `<link rel="canonical">` on every page to the mathlib4 docs root. The previous `canonical_roots` mapping was self-referential (mathlib3 → mathlib3), which is a no-op for SEO.

The motivation is that the deprecated mathlib3 docs consistently outrank the mathlib4 docs in Google and DuckDuckGo, which is confusing for newcomers. A blanket canonical pointer to the mathlib4 docs root is a deliberately simple choice — a per-page mapping was considered (snake_case → CamelCase covers ~67% of files cleanly) but rejected as not worth the complexity given that many files have been split, renamed, or moved.

Discussion: https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/De-emphasizing.20mathlib3.20docs.20in.20search.20results.20via.20canonical

🤖 Prepared with Claude Code


bryangingechen:
#deploy

bryangingechen:
#deploy

bryangingechen:
Screenshot 2026-05-07 at 12 49 01 AM
Here's what a typical page looks like. I think `display: flex` is causing there to be missing spaces between some of the text elements and the links. Should be easy to fix.

templates/base.j2
```diff
@@ -22,6 +22,13 @@
 
 <body>
 
+<aside class="deprecated_banner" role="alert">
+    <strong>mathlib3 is deprecated.</strong>
+    These docs are for the Lean 3 version of mathlib, which is no longer maintained.
+    For the active Lean 4 library, see the
+    <a href="https://leanprover-community.github.io/mathlib4_docs/">mathlib4 docs</a>.
+</aside>
```
FordUniver, line 29:
```suggestion
<aside class="deprecated_banner" role="alert">
    <strong>mathlib3 is no longer maintained</strong> &mdash; see the
    <a href="https://leanprover-community.github.io/mathlib4_docs/">mathlib4 docs</a>
    for the current library.
</aside>
```

style.css
```diff
@@ -107,6 +108,32 @@ label[for="nav_toggle"] {
...
+    color: white;
+    font-size: 0.95em;
+    line-height: 1.3;
+    text-align: center;
+    padding: 0 1em;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    z-index: 2;
+}
```
FordUniver, line 120:
```suggestion
.deprecated_banner {
    position: fixed;
    top: 0; left: 0; right: 0;
    min-height: var(--banner-height);
    padding: 0.4em 1em;
    background: #b00020;
    color: white;
    text-align: center;
    z-index: 2;
}
```

FordUniver (changes requested):
I think this scales cleanly across screen sizes

print_docs.py
```diff
@@ -252,20 +252,13 @@ def convert_markdown(ds):
...
+# maintained library.
+MATHLIB4_DOCS_ROOT = 'https://leanprover-community.github.io/mathlib4_docs/'
 
 def get_canonical_url(path, project='mathlib'):
-  try:
-    root = canonical_roots[project]
-  except KeyError:
-    return None
-  return root + '/' + path
+  return MATHLIB4_DOCS_ROOT
```
FordUniver: I think blanket pointing to the root of mathlib4 docs is exactly the thing that will lead search engines to ignore a canonical link. There needs to be some reasonable overlap in content, so the map is needed if we do the canonical link approach instead of just a de-index.

FordUniver: I think the best approach is to commit a one-off snapshot [a map that covers 90%+ of the mathlib3 docs with newer (current) mathlib4 references](https://github.com/FordUniver/doc-gen/blob/deprecate-banner-and-canonical/mathlib4_canonical_map.yaml). The code could then just check if a replacement is known:

```python
import yaml as _yaml

MATHLIB4_DOCS = 'https://leanprover-community.github.io/mathlib4_docs/'
try:
  with open(os.path.join(os.path.dirname(__file__), 'mathlib4_canonical_map.yaml')) as _f:
    _mathlib4_map = _yaml.safe_load(_f) or {}
except FileNotFoundError:
  _mathlib4_map = {}

def get_canonical_url(path, project='mathlib'):
  if project == 'mathlib' and path.endswith('.html'):
    target = _mathlib4_map.get(path[:-5].replace('/', '.'))
    if target:
      return MATHLIB4_DOCS + target + '.html'
  try:
    root = canonical_roots[project]
  except KeyError:
    return None
  return root + '/' + path
```

FordUniver (changes requested):

print_docs.py
bryangingechen: I like this idea. Do you mind making a PR into this branch?

FordUniver: > I like this idea. Do you mind making a PR into this branch?

Done, see #181

FordUniver: @kim-em Have you had a chance to look at the PR yet?

PR #181, opened 2026-05-07
feat: per-page canonical via mathlib3→mathlib4 map

FordUniver:


bryangingechen (approved):
OK, makes sense! @kim-em: what do you think?

